PR #3, opened 2018-10-12
Don't keep Cargo.lock in source control.

waywardmonkeys:
Since this is a library, having the `Cargo.lock` in source control isn't typical / needed.


matklad:
.gitignore should be updated I guess :)

waywardmonkeys:
Updated `.gitignore`.

PR #4, opened 2018-10-31
Add wrap_internal function

jD91mZM2:
Needed for parsing `a + b` as `Add(a, b)`, since you *first* need to parse `a` and *then* maybe wrap it.

### DIFF
--- a/src/green.rs
+++ b/src/green.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use {SmolStr, TextUnit, Types};
+use crate::{SmolStr, TextUnit, Types};
 
 /// `GreenNode` is an immutable syntax tree,
 /// which is cheap to update. It lacks parent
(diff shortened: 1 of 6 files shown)


src/green.rs
```diff
@@ -58,6 +62,31 @@ impl<T: Types> GreenNodeBuilder<T> {
...
+    ///   self.parse_expr();
+    ///   builder.finish_internal();
+    /// }
+    /// ```
+    pub fn wrap_checkpoint(&self) -> WrapCheckpoint {
+        WrapCheckpoint(self.children.len())
+    }
+    /// Wrap the previous branch marked by wrap_checkpoint in a new branch and
+    /// make it current.
+    pub fn wrap_internal(&mut self, checkpoint: WrapCheckpoint, kind: T::Kind) {
```
matklad: Nifty!

I thought one needs to create a wrapper around green node to track such state (I do this in rust-analyzer), but looks like this indeed work!

Lets name it `start_internal_at(checkpoint, kind)`, to make it more obvious that `start _internal + start_internal_at = finish_internal`.

Also, I think there's one more invariant we need to check: the `fist_child` fields of `parents` should be increasing.

So, we need to check that top's first_child is <= checkpoint. I think? Not sure about this one.


It would be awesome to exercise this by writing a tests which builds a tree for `1 + 2 * 3  + 4` (without the parser, just calling the builder).

jD91mZM2: I'm using this in rnix currently and it seems to work

jD91mZM2:
@matklad Should `wrap_checkpoint` also be renamed?

matklad:
Yeah, let's rename it ()and the struct) to just checkpoint?

PR #10, opened 2019-01-11
rename TreePtr -> TreeArc

matklad:
It's the same amount of characters but gives much more information:

`TreePtr` could behave like a Box or Rc or Arc. TreeArc is clear about
semantics.

@jD91mZM2 what do you think about this naming?

### DIFF
--- a/examples/math.rs
+++ b/examples/math.rs
@@ -48,7 +48,7 @@ impl rowan::Types for Types {
 }
 
 type Node = rowan::SyntaxNode<Types>;
-type TreePtr<T> = rowan::TreePtr<Types, T>;
+type TreeArc<T> = rowan::TreeArc<Types, T>;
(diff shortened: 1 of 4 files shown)


jD91mZM2:
I like it better definitely! Although I feel like going more clear technically might not be the right path. Maybe the *intent* should be clearer instead: Such as `Owned<SyntaxNode>`. Just a suggestion.

matklad:
`Owned` seems too generic: it's not clear that you can apply it only to tree stuff, and `Box` and `Arc` are both *Owned*, but are pretty different. 

I am sort-of torn on "intent" vs "implementation": often times, the implementation is the most succinct way to describe intent: `Arc` vs `CheaplyCopyableThreadSafeSharedOwndership`

PR #11, opened 2019-03-08
Add de/serialization capability

CAD97:
Add a simple "obviously correct" serde de/serialization impl behind the "serde" feature.

A node is represented as a singleton map from the syntax kind to an anonymous enumeration of either the leaf text or the branch's children. Due to the use of the anonymous enumeration, only self-describing formats can be deserialized from.

I suspect that this will be most useful for serializing to textual formats for comparison in tests, similar to the debug representation. (In fact, I have an adapter doing so in my usage.) Using a structured representation carries some benefit (editor folding support and such) and theoretically can enable more semantic diffing than a textual diff.

Omitting span information from the representation doesn't lose information as this representation instead represents the leaf text. This also means that inserting a new subtree doesn't change the representation before or after it, thus reducing the size of the diff.

Serialization and Deserialization are provided in the same file to help the obviousness of the implementation. Deserialization is mostly along for the ride with the Serialization, but could be dropped if not desired to be supported, as it does add significant complexity compared to the Serialize implementation.

An untagged enum is handled via serde_derive to avoid reimplementing that complexity locally, and to make it more likely to be correct.

-----

(needs tests)

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1,0 +1,86 @@
+use crate::*;
+use serde::{
+    de::{Deserializer, Error, MapAccess, Visitor},
+    ser::{SerializeMap, Serializer},
+    Deserialize, Serialize,
+};
+use std::{fmt, marker::PhantomData};
+
+///////////////////////////////////////////// ~ser~~~ /////////////////////////////////////////////
+
+#[derive(Serialize)]
+#[serde(untagged)]
+#[serde(bound(serialize = "T::Kind: Serialize"))]
+enum NodeContentSer<'a, T: Types> {
+    Leaf(&'a str),
+    Branch(&'a [GreenNode<T>]),
+}
+
+impl<T: Types> Serialize for GreenNode<T>
+where
+    T::Kind: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(1))?;
+        if let Some(text) = self.leaf_text() {
+            map.serialize_entry(&self.kind(), &NodeContentSer::<T>::Leaf(text.as_str()))?;
+        } else {
+            map.serialize_entry(&self.kind(), &NodeContentSer::Branch(self.children()))?;
+        }
+        map.end()
+    }
+}
+
+///////////////////////////////////////////// ~~~~de~ /////////////////////////////////////////////
+
+#[derive(Deserialize)]
+#[serde(untagged)]
+#[serde(bound(deserialize = "T::Kind: Deserialize<'de>"))]
+enum NodeContentDe<'a, T: Types> {
+    Leaf(&'a str),
+    Branch(Vec<GreenNode<T>>),
+}
+
+struct NodeVisitor<T: Types>(PhantomData<GreenNode<T>>);
+
+impl<'de, T: Types> Visitor<'de> for NodeVisitor<T>
+where
+    T::Kind: Deserialize<'de>,
+{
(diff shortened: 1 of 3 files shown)


matklad:
I think I am fine with it, but I’d love to hear more about motivation. 

If the sole goal here is serialization to textual format for comparison, than I think walking the tree manually with postorder should be more convenient than going via serde.

matklad:
Not that the serialization format is a public API, so changing the way we represent nodes will be a breaking change.

src/ser.rs
```diff
@@ -0,0 +1,86 @@
...
+where
+    T::Kind: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(1))?;
+        if let Some(text) = self.leaf_text() {
+            map.serialize_entry(&self.kind(), &NodeContentSer::<T>::Leaf(text.as_str()))?;
```
dtolnay:
```suggestion
            map.serialize_entry(&self.kind(), text.as_str())?;
```

src/ser.rs
```diff
@@ -0,0 +1,86 @@
...
+        } else {
+            map.serialize_entry(&self.kind(), &NodeContentSer::Branch(self.children()))?;
```
dtolnay:
```suggestion
            map.serialize_entry(&self.kind(), self.children())?;
```

src/ser.rs
CAD97: I assume these are equivalent then? I originally used the `#[serde(untagged)]` enum for obvious-correctness.

dtolnay: Yes, untagged newtype variants behave like the underlying type i.e. the NodeContentSer derive expands to:

```rust
impl<'a, T: Types> Serialize for NodeContentSer<'a, T>
where
    T::Kind: Serialize,
{
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        match self {
            NodeContentSer::Leaf(leaf) => Serialize::serialize(leaf, serializer),
            NodeContentSer::Branch(branch) => Serialize::serialize(branch, serializer),
        }
    }
}
```

src/ser.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    fn visit_map<A>(self, mut map: A) -> Result<GreenNode<T>, A::Error>
+    where
+        A: MapAccess<'de>,
+    {
+        let kind: T::Kind = map
+            .next_key()?
+            .ok_or_else(|| A::Error::custom("unexpected empty map"))?;
+        let value: NodeContentDe<T> = map.next_value()?;
```
dtolnay: Untagged enums generally have pretty poor error reporting on deserialization. If there was any type error somewhere inside the vec of GreenNode, all we can know is that the input "Isn't" a vec of GreenNode so you get a message that none of the NodeContentDe variants were matched, rather than pinpointing the problem. I don't know how much you care about this but it would be possible to do better by handwriting the Deserialize impl for NodeContentDe to have a visit_str method and visit_seq method.

CAD97:
I'm dropping this PR: it's a lot less useful now that the node types are not-generic. I'm still + on that change, but it means that this has to be done by the consumer if it's wanted.

The reason to use serde rather than a manual postorder walk (or `.to_string()`) is that I want to be able to work with an actual data format that I can manipulate rather than something ad-hoc.

matklad:
Added serialization in https://github.com/rust-analyzer/rowan/pull/27

PR #12, opened 2019-03-28
Rewrite to separate tokens from interior nodes

matklad:
This PR rewrites the API of rowan once again, to better handle leaf nodes (tokens).

Specifically:

* leaf tokens are now stored inline (no separate allocation) in green nodes, and are not stored in red nodes at all (this should help with memory usage, as tokens tend to occupy more than a half of the tree)
* separate `Token` and `Node` types: one has children, another has text

### DIFF
--- a/src/algo.rs
+++ b/src/algo.rs
@@ -1,0 +1,256 @@
+use std::ops::Range;
+
+use crate::{Types, generate, SyntaxNode, SyntaxElement, TextUnit, TextRange, SyntaxToken, SyntaxIndex};
+
+/// `WalkEvent` describes tree walking process.
+#[derive(Debug, Copy, Clone)]
+pub enum WalkEvent<T> {
+    /// Fired before traversing the node.
+    Enter(T),
+    /// Fired after the node is traversed.
+    Leave(T),
+}
+
+/// There might be zero, one or two leaves at a given offset.
+#[derive(Clone, Debug)]
+pub enum TokenAtOffset<T> {
+    /// No leaves at offset -- possible for the empty file.
+    None,
+    /// Only a single leaf at offset.
+    Single(T),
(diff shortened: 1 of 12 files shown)


src/green.rs
```diff
@@ -2,65 +2,170 @@ use std::{mem::size_of, sync::Arc};
...
+        &self.children
+    }
+
+    /// Number of memory bytes of occupied by subtree rooted at `self`.
+    pub(crate) fn memory_size_of_subtree(&self) -> usize {
+        let mut res = size_of::<Self>();
+        self.children().iter().for_each(|el| match el {
+            GreenElement::Token(token) => {
+                res += size_of::<GreenToken<T>>();
+                if token.text.len() > 22 {
```
kjeremy: 22?

matklad: Yep, that”s the length when we start to heap-allocat small strings

kjeremy: Can that be thrown into a constant somewhere?

kjeremy:
LGTM

src/algo.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn left_biased(self) -> Option<T> {
+        match self {
+            TokenAtOffset::None => None,
+            TokenAtOffset::Single(node) => Some(node),
+            TokenAtOffset::Between(left, _) => Some(left),
+        }
+    }
+}
+
+impl<T> Iterator for TokenAtOffset<T> {
```
kjeremy: Looks like you could implement `ExactSizeIterator` here but it's probably not worth it.

src/green.rs
jD91mZM2: Or a function in SmolStr, `is_inline()`

matklad: good call!

PR #20, opened 2019-07-04
Update docs

jDomantas:
* update docs for `GreenNodeBuilder::checkpoint` and make its code sample into a non-ignored doc test
* update other docs and panic messages that referred to `start_internal` and `finish_internal`
* remove `.into()`s that are no longer needed after the last change to `s_expressions.rs` example


matklad:
awesome, thanks!

bors r+

matklad:
bors ping

PR #21, opened 2019-07-06
Remove unneeded lifetime

kjeremy:


src/algo.rs
```diff
@@ -150,9 +150,9 @@ impl SyntaxNode {
     /// Traverse the subtree rooted at the current node (including the current
     /// node) in preorder, including tokens.
     #[inline]
-    pub fn preorder_with_tokens<'a>(
-        &'a self,
-    ) -> impl Iterator<Item = WalkEvent<SyntaxElement<'a>>> {
+    pub fn preorder_with_tokens(
+        &self,
+    ) -> impl Iterator<Item = WalkEvent<SyntaxElement>> {
```
matklad: let's use `'_`

matklad:
We now removed all lifetimes everywhere :D

PR #22, opened 2019-07-08
Cursor

matklad:
It's been a long time since I've re-written the core of rowan, so I figure out this is a high time for yet another rewrite! :) 

The main motivation here is memory usage: rowan trees weigh a ton. Specifically, it is SyntaxNodes (red trees) which are heavy:

```
GreenNode    24 bytes
SyntaxNode   72 bytes
```

An [old thread](https://github.com/KirillOsenkov/Bliki/wiki/Roslyn-Immutable-Trees) about roslyn is illuminating here (thanks @KirillOsenkov for documenting this!). The way they solve this problem in roslyn is by using weak references for some red-tree links (in particular, function bodies are stored using weak refs). This allows GC to collect red trees while maintaining reference identity, which is super cool.

Unfortunately, we can't easily use weak refs in rust: the central property of the previous solution was that traversal always yields a plain reference, and we can't track if a plain reference is used or not. 

So, this PR proposes a radically different approach for handing `SyntaxNode`s. Instead of lazily allocating red nodes in arena, we create and tear down them on the fly during traversal. That is, at every instant during traversal we only keep in memory the red path to the root. Naively, this can be implemented as

```rust
struct SyntaxNode {
    parent: Arc<SyntaxNode>,
    offset: TextUnit,
    green: *const GreenNode,
}
```

That would be pretty inefficient though: traversal would require a ton of allocations and atomic reference count increments! 

So we do something horrible and beautiful instead: we make `SyntaxNode`s not `Send/Sync`. This allows us to:

* use `Rc` instead of an `Arc`
* store a thread-local free list of `SyntaxNode`s

`!Send` seems strange and limiting, but should work fine in practice I think: the green tree itself is perfectly thread safe still, it's only a particular tree traversal that needs to happen on a specific node. If you want to send a particular node to the other thread, you can send the root node together with a "pointer" (text span) inside the tree. 

On the other hand, traversal becomes:

* access thread local (and we can even get rid of this by caching reference to free list in the node itself)
* bump rc
* push/pop a node off a free list

This still seems costlier than our previous reference-based traversal, but I am not sure that's the case without the benchmarks:

* the working set is smaller: due to free list, we basically reused the same nodes for traversal, instead of walking through memory
* there are no atomic operations at all

Benchmarking this properly will require adapting rust-analyzer to the new API :)

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -1,0 +1,734 @@
+use std::{
+    slice, ptr, iter, mem,
+    rc::Rc,
+    marker::PhantomData,
+    cell::{Cell, RefCell},
+    hash::{Hash, Hasher},
+};
+
+use crate::{
+    GreenNode, GreenElement, TextUnit, TextRange, GreenToken, SyntaxKind, SmolStr, WalkEvent,
+    TokenAtOffset,
+};
+
+#[derive(Debug, Clone)]
+pub struct SyntaxNode(Rc<NodeData>);
+
+impl Drop for SyntaxNode {
+    fn drop(&mut self) {
+        NodeData::delete(&mut self.0)
+    }
+}
+
+impl PartialEq for SyntaxNode {
+    fn eq(&self, other: &SyntaxNode) -> bool {
+        ptr::eq(self.green(), other.green())
+            && self.text_range().start() == other.text_range().start()
+    }
+}
+
+impl Eq for SyntaxNode {}
+
+impl Hash for SyntaxNode {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        ptr::hash(self.green(), state);
+        self.text_range().start().hash(state);
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SyntaxToken {
+    parent: SyntaxNode,
+    index: u32,
+    offset: TextUnit,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum SyntaxElement {
+    Node(SyntaxNode),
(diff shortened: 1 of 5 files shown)


src/cursor.rs
```diff
@@ -0,0 +1,366 @@
...
+
+struct FreeList {
+    first_free: Option<Rc<NodeData>>,
+    len: usize,
+}
+
+const FREE_LIST_MAX_LEN: usize = 128;
+
+thread_local! {
+    static FREE_LIST: FreeList = FreeList::new()
```
pcpthm: this variable is unused

jD91mZM2:
I've really only used the `SyntaxNode` kind, so I think if we're removing `Send` + `Sync` from there we'll need a way to convert it back to a `GreenNode` and back. Otherwise, I love the idea of using `Rc` for most computation as `Arc`s are pretty expensive :)

matklad:
preliminary testing (parsing parser.rs and debug-dumping it to dev/null) show that this is indeed significantly more memory efficient and slightly faster

matklad:
perf measuremets in https://github.com/rust-analyzer/rust-analyzer/pull/1545

matklad:
>I've really only used the SyntaxNode kind, so I think if we're removing Send + Sync from there we'll need a way to convert it back to a GreenNode and back. 

Curiously, in the whole of rust-analyzer this was ever needed for root nodes, so a type which stores syntax errors as well is used for this: https://github.com/rust-analyzer/rust-analyzer/blob/cf932181cf18c1e5fa99cb024bdb92784ed9649e/crates/ra_syntax/src/lib.rs#L60-L65

PR #24, opened 2019-07-19
Make using rowan as a library nicer

matklad:


src/api.rs
```diff
@@ -1,10 +1,403 @@
...
-    pub fn ancestors(&self) -> impl Iterator<Item = SyntaxNode> {
-        std::iter::successors(Some(self.clone()), SyntaxNode::parent)
+pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
+    fn debug_kind(kind: &SyntaxKind<Self>, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&kind.raw, f)
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SyntaxKind<L> {
```
jD91mZM2: I thought we agreed on [using traits which the user could implement themselves?](https://gitlab.com/jD91mZM2/rnix/merge_requests/10#note_175424408)

src/syntax_text.rs
```diff
@@ -0,0 +1,223 @@
+use std::{fmt, ops};
+
+use crate::{
+    cursor::{SyntaxElement, SyntaxNode},
+    SmolStr, TextRange, TextUnit,
+};
+
+#[derive(Clone)]
+pub struct SyntaxText {
```
jD91mZM2: This might need some docs because I have no idea what it is

examples/s_expressions.rs
```diff
@@ -16,18 +15,43 @@ use rowan::SmolStr;
...
-const L_PAREN: SyntaxKind = SyntaxKind(0); // '('
-const R_PAREN: SyntaxKind = SyntaxKind(1); // ')'
-const WORD: SyntaxKind = SyntaxKind(2); // '+', '15'
-const WHITESPACE: SyntaxKind = SyntaxKind(3); // whitespaces is explicit
-const ERROR: SyntaxKind = SyntaxKind(4); // as well as errors
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum Lang {}
+impl rowan::Language for Lang {
+    fn debug_kind(kind: &SyntaxKind, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let tag = match *kind {
```
jD91mZM2: I definitely *definitely* want a macro generating this. Not procedural ones or anything too fancy, just
```rust
rowan_enum! {
    onto Lang,
    ---
    L_PAREN = 0,
    R_PAREN, // implicit = 1 and so on
}
```
=>
```rust
const L_PAREN: SyntaxKind = SyntaxKind::new(0);
const R_PAREN: SyntaxKind = SyntaxKind::new(1);

impl rowan::Language for Lang {
    fn debug_kind(kind: &SyntaxKind, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{}", match *kind {
            // actually converting an identifier to a string with a
            // macro is not well-defined, but for such very simple
            // cases as this one I think one can't possibly blame us
            // if we rely on `stringify!`
            L_PAREN => "L_PAREN",
            R_PAREN => "R_PAREN",
        })
    }
}
```

src/api.rs
matklad: Yeah, I've tried that approach, and it turned out to be pretty unwieldy. 

The biggest problem is that the trait-based approach needs a swamp of associated types, and associated types don't work with `impl Trait` (so, using `std::iter::successors` becomes impossible) and they also don't map nicely to `SyntaxElement` enum.

The second problem is that you can't actually use the trait approach to write blanket impls for Debug & Display. You can supply an utility function `debug_fmt`, but the user will have to write an `impl`.

The third problem is that the user will need to import API traits to use them, instead of all methods being inherent.

So I want to try this "non-template base class" pattern, but via composition, and I want to try to migrate rust-analyzer to it, instead of manually doing forwarding there. I *think* my desire to be able to write trait impls is an instance of YAGNI

jD91mZM2: I see, thanks for the insight! :+1: for this way, then

matklad: also cc https://users.rust-lang.org/t/api-design-question-how-to-wrap-a-family-of-types/30424

src/syntax_text.rs
matklad: Yeah, the docs are yet nonexistent, will add them.

`SyntaxText` is a "string" type. It represents a text of a single syntax node. It can be useful to do things like "how many newlines are there in this node"?

examples/s_expressions.rs
matklad: sgtm!

src/syntax_text.rs
jD91mZM2: Ah, it recurses children?

matklad: yep!

src/api.rs
jD91mZM2: Crazy idea: A macro to generate code similar to Rust analyzer, that wraps a SyntaxNode. Not using traits will allow you to use `impl Trait`. It won't be pretty, but... I guess it'll let users implement their own `Debug`...........

jD91mZM2: Or, what about abusing that root data field? Since you're already using dynamic dispatch to avoid generics, what if we change `Option<Box<dyn Any + Send + Sync>>` to `Box<dyn Lang>` where
```rust
trait Lang: Any + Send + Sync {
    fn debug_kind(...)
}
// To store root data, one now adds a field to their own struct implementing `Lang`
```

matklad: Root data is gone in this version :) But yeah, I imagine using dynamic dispatch here *might* work

examples/s_expressions.rs
matklad: Heh, I had made Kind an associated type: in rust-analyer, kinds live in a crate which doesn't know about the sytanax tree at all.

matklad:
rust-analyzer integration: https://github.com/rust-analyzer/rust-analyzer/pull/1570

examples/s_expressions.rs
jD91mZM2: That works, then I could switch back to an enum I guess

PR #33, opened 2019-10-30
Use variable length encoding variable length array for GreenNode

CAD97:
```
ArcGreenNode      8
GreenToken        32
OwnedGreenElement 40
GreenElementRef   16

SyntaxNode    8
SyntaxToken   16
SyntaxElement 24
```

The layout of a GreenNode:

```
// align(u64)
head: {
    text_len: TextUnit (u32)
    kind: SyntaxKind (u16)
    child_count: u16
}
tail: [u64; dyn]
// 0usize, ArcGreenNode (1 usize tag (padded to u64), 1 usize data (padded to u64))
// 1usize, GreenToken (1 usize tag (padded to u64), 4 u64 data)
```

`ArcGreenNode` is required so that `Drop` does the right thing for `GreenNode`. `GreenNode` is currently `Sized` according to `rustc`, and should probably be made `?Sized` via making the tail an `extern type` once that is stable.

The code quality is very proof-of-concept. I'm fairly confident it does the right thing, but it could use some review and proper organization before production use.


src/cursor.rs
```diff
@@ -800,15 +797,18 @@ impl GreenNode {
...
-            (element, (end_index - index - 1, offset))
-        })
+    ) -> impl Iterator<Item = (GreenElementRef<'_>, (usize, TextUnit))> {
+        // FIXME: uhoh: reverse iteration?
+        //        self.children().take(end_index).rev().enumerate().map(move |(index, element)| {
+        //            offset -= element.text_len();
+        //            (element, (end_index - index - 1, offset))
+        //        })
+        unimplemented!();
+        vec![].into_iter()
```
CAD97: @matklad, You suggested the variable length encoding, so perhaps you can suggest the best way to deal with this.

`children_to` is used in `last_child`, `last_child_or_token`, `prev_sibling`, and `prev_sibling_or_token`. If we do go with this layout, either the cursor has to keep a previous pointer or previous becomes a decently more expensive operation.

src/cursor.rs
```diff
@@ -565,7 +563,7 @@ impl SyntaxToken {
     }
 
     pub fn green(&self) -> &GreenToken {
-        self.parent.green().children()[self.index as usize].as_token().unwrap()
+        self.parent.green().children().nth(self.index as usize).unwrap().as_token().unwrap()
```
CAD97: This should probably be adjusted to be a direct access.

src/cursor.rs
```diff
@@ -788,8 +785,8 @@ impl GreenNode {
         &self,
         start_index: usize,
         mut offset: TextUnit,
-    ) -> impl Iterator<Item = (&GreenElement, (usize, TextUnit))> {
-        self.children()[start_index..].iter().enumerate().map(move |(index, element)| {
+    ) -> impl Iterator<Item = (GreenElementRef<'_>, (usize, TextUnit))> {
+        self.children().skip(start_index).enumerate().map(move |(index, element)| {
```
CAD97: This should probably be adjusted to be a direct offset.

src/cursor.rs
```diff
@@ -800,15 +797,18 @@ impl GreenNode {
...
         end_index: usize,
         mut offset: TextUnit,
-    ) -> impl Iterator<Item = (&GreenElement, (usize, TextUnit))> {
-        self.children()[..end_index].iter().rev().enumerate().map(move |(index, element)| {
-            offset -= element.text_len();
-            (element, (end_index - index - 1, offset))
-        })
+    ) -> impl Iterator<Item = (GreenElementRef<'_>, (usize, TextUnit))> {
+        // FIXME: uhoh: reverse iteration?
+        //        self.children().take(end_index).rev().enumerate().map(move |(index, element)| {
```
CAD97: This should probably be adjusted to be a direct offset.

CAD97:
TL;DR: this is even messier than #32 but has a better chance at improving memory usage at least.

CAD97:
I put TODOs inline for where I mentioned maybe switching to direct FAM indexing. I also added a silly workaround for the reverse iteration to restore functionality without changing the interface.

src/cursor.rs
matklad: Hm, would a *single* previous pointer actually help in this case? You'd be able to go one step backwards, but then you wouldn't have previous-previous pointer. 

I think we should store a byte-sized tag between the elements, which describes prev/next elements. *maybe* we should even use XOR-linked list trick for the tag, but I doubt it'll be actually important.

CAD97: The tag is necessarily `u64` big for alignment, so we've a good deal of space to stick extra info. If I'm reading you right, then instead of the tag being 0u64 and 1u64, it should be 0u32 and 1u32, but we also provide a tag for the previous kind to allow for reverse iteration. (Or equivalently any other packing of two bits into 64.)

That's actually pretty clever and would work perfectly!

If we wanted to get crazy each tag could store 32 (if we want an end marker) or even 64 kinds for larger strides, but that definitely seems like overkill.

CAD97: Found an issue while trying to implement it: we need a way to find the end to start with. And the header of the `GreenNode` is full up with `AtomicUsize, TextUnit(u32), SyntaxKind(u16), child_count(u16)`.

We can give up `ExactSizeIterator` and store `fam_length(u16)` instead to allow reverse iteration. I'm going to implement that approach.

CAD97: Ugh, right, we can't have `.rev().enumerate()` without `DoubleEndedIterator + ExactSizeIterator`.

Time for me to examine what's actually being done and try to find a way around wanting both `DoubleEndedIterator` _and_ `ExactSizeIterator`.

CAD97:
I hooked this up to ra and got

> test result: FAILED. 2 passed; 198 failed; 0 ignored; 0 measured; 0 filtered out

so something's messed up, obviously.

CAD97:
I don't really have the resources to debug this, and the boxed/deduped tokens approach will hopefully make this unnecessary, so I'm going to let this PR sit idle for a bit while I try out that approach.

CAD97:
Closing this as the approach of boxing tokens makes this code much simpler and less error-prone.

PR #34, opened 2019-11-02
Boxed tokens

CAD97:
This is the result of applying [this technique](https://github.com/rust-analyzer/rowan/pull/32#issuecomment-548603444) of boxing/deduping `GreenToken` to compress `GreenElement` and, as a result, the allocation for a `GreenNode`'s children. It also applies a minimal flexible array member technique for the `GreenNode`'s children, allocating them inline.

This is as far as the technique can go with using the standard `Arc` and `NodeOrToken`. `GreenElement = NodeOrToken<Arc<GreenNode>, Arc<GreenToken>>` is 3x`usize` large, but [can theoretically be niched to 2x`usize`](https://github.com/rust-lang/rust/issues/66029). (NB: the same niche optimization can be applied to `SyntaxElement`.)

I've split this PR up into commits each representing a logical step.

```
GreenToken      32
Arc<GreenNode>  16
Arc<GreenToken> 8
GreenElement    24

SyntaxNode    8
SyntaxToken   16
SyntaxElement 24
```

It is possible to optimize size further, but I'd like to propose those in a second PR on top of this one. Remaining potential optimization steps along these lines:

- Implement a specific `GreenElement` to manually niche `Arc<GreenNode>` and `Arc<GreenToken>`.
    - [Could be done by the compiler.](https://github.com/rust-lang/rust/issues/66029) `sizeof(GreenElement)` = 2x`usize`.
- Implement a specific `SyntaxElement` to manually niche `SyntaxNode` and `SyntaxToken`.
    - [Could be done by the compiler.](https://github.com/rust-lang/rust/issues/66029) `sizeof(SyntaxElement)` = 2x`usize`.
- Implement a specific `Arc<GreenNode>` to make ptr-to-`GreenNode` a thin ptr instead of a fat ptr.
    - Standard `GreenElement = NodeOrToken` would be 2x`usize`. Manually non-zero-cost niched `GreenElement` (tag in alignment bits) would be 1x`usize`.

Results in rust-analyzer:

<details><summary>With this branch</summary>

```powershell
PS D:\usr\Documents\Code\Rust\rust-analyzer> cargo run --bin ra_cli --release -- analysis-stats .
    Finished release [optimized + debuginfo] target(s) in 0.46s
     Running `target\release\ra_cli.exe analysis-stats .`
Database loaded, 221 roots, 1.0601376s
Crates in this dir: 27
Total modules found: 331
Total declarations: 11135
Total functions: 3839
[...]

PR #35, opened 2019-11-05
Custom GreenElement

CAD97:
Implements a thin-pointer version of `GreenNode` and uses that to implement a 1×`usize` `GreenElement`. We do not yet apply any manual niching to `SyntaxElement`, though it still could [be niched to 2×`usize`](https://github.com/rust-lang/rust/issues/66029) at zero performance cost.

We provide a `match_element!` macro that allows for (what looks like) pattern matching over `GreenElement` even though it is no longer an enum.

```
GreenToken      32
ArcGreenNode    8
Arc<GreenToken> 8
GreenElement    8

SyntaxNode    8
SyntaxToken   16
SyntaxElement 24
```

This passes miri! 🎉 We should probably add miri to CI to make sure it stays that way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ rustc-hash = "1.0.1"
 text_unit = "0.1.6"
 smol_str = "0.1.10"
 serde = { version = "1.0.89", optional = true, default-features = false }
-static_assertions = "1.0.0"
 
 [dev-dependencies]
 m_lexer = "0.0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ rustc-hash = "1.0.1"
 text_unit = "0.1.6"
 smol_str = "0.1.10"
 serde = { version = "1.0.89", optional = true, default-features = false }
+thin-dst = "1.0.0-beta.4"
 
 [dev-dependencies]
 m_lexer = "0.0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ rustc-hash = "1.0.1"
 text_unit = "0.1.6"
 smol_str = "0.1.10"
 serde = { version = "1.0.89", optional = true, default-features = false }
+static_assertions = "1.0.0"
 
 [dev-dependencies]
 m_lexer = "0.0.4"
--- a/examples/math.rs
+++ b/examples/math.rs
@@ -106,7 +106,7 @@ impl<I: Iterator<Item = (SyntaxKind, SmolStr)>> Parser<I> {
         self.parse_add();
         self.builder.finish_node();
 
-        SyntaxNode::new_root(self.builder.finish())
+        SyntaxNode::new_root(self.builder.finish().unwrap_node())
     }
 }
 
--- a/examples/math.rs
+++ b/examples/math.rs
@@ -111,7 +111,7 @@ impl<I: Iterator<Item = (SyntaxKind, SmolStr)>> Parser<I> {
 }
 
 fn print(indent: usize, element: SyntaxElement) {
-    let kind: SyntaxKind = element.kind().into();
+    let kind: SyntaxKind = element.kind();
     print!("{:indent$}", "", indent = indent);
     match element {
         NodeOrToken::Node(node) => {
--- a/examples/s_expressions.rs
+++ b/examples/s_expressions.rs
@@ -58,9 +58,7 @@ type SyntaxElement = rowan::NodeOrToken<SyntaxNode, SyntaxToken>;
 
 /// GreenNode is an immutable tree, which is cheap to change,
 /// but doesn't contain offsets and parent pointers.
-use rowan::GreenNode;
-
-use std::sync::Arc;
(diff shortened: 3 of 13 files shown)


CAD97:
https://github.com/rust-analyzer/rust-analyzer/pull/2182 confirms that this decreases rowan memory usage, with `ParseQuery` dropping from 30MB on master, to 18MB with #34, and 9MB with this patch.

The latest commit returns `GreenNode` to being actually `!Sized`, but keeps the benefit of thin pointers in `ArcGreenNode` and `GreenElement` by storing type-erased thin pointers and recovering the length from the header.

This passes `cargo miri test` now 🎉 We should probably run miri in CI to make sure it stays that way.

@matklad I believe this is ready for review now. I think we should leave the [`enum NodeOrToken` niche optimization](https://github.com/rust-lang/rust/issues/66029) to the compiler to eventually pick up; we'd want to revert to the safe code once it does anyway, so I don't think it's worth the short term code clarity cost to manually take advantage of that niche.

CAD97:
Review hint: each commit should be self contained _except a392206 and 1f42bcb_, which need to be taken together. Review will probably be easier piecewise by commit (except for that one exception).

matklad:
@CAD97 did you by any change do any measurements of how effective the interning is? Ie, if we parse something like the old, non-split version of parser.rs from rustc, how many tokens and nodes are there, and which proportion of tokens is deduped? "30MB >18MB > 9MB" size reduction looks super sweet, up to the point it's hard to believe in :)

matklad:
It's also interesting that `SmloStr` now becomes completely unnecessary: if we heap allocate tokens anyway, it's better to store token text inline, using the trailing objects trick h.

CAD97:
This now uses thin-dst. It could use some cleanup, but it also fully encapsulates `ArcGreenNode` and thin-dst; the outside world only sees the std Arc.

CAD97:
Back-of-the-envelope space savings calculation:

`GreenElement` went from 40 bytes to 8 bytes for a saving of 32 bytes per element. As a very rough underestimate of the number of nodes seen by `ra analysis-stats ra`, modules + declarations + functions + expressions = 104,548 nodes. That's 3,345,536 bytes saved, or 3.2MiB. For every non-deduped token, we go -8 on bytes allocated, as we go from a 40 byte `GreenElement` with the token inline, to a 8 byte `GreenElement` pointer to a 48 byte `ArcInner<GreenToken>`. For each deduped token, we save 32 bytes. Every node has at least one token, and statistically this severe underestimation is entirely deduped. So that's another 3.2MiB saved.

So an extremely conservative low-effort estimation puts this at at least a 20% memory savings. For a better estimation, we should hook ra analysis-stats to print an estimation of the full node count. A quick way would be to just annotate the [cache](https://github.com/CAD97/rowan/blob/2ce4c6be75897f9c45cc9cba95e730e75294142d/src/green/builder.rs#L20-L48) to log cache hits and misses.

So, @matklad, if this all looks good to you, I'll publish version 1.0.0 of thin-dst (transfer ownership to the rust-analyzer org?) and bump this PR to depend on that so we can start realizing these benefits.

CAD97:
It'd still be cool to make `GreenToken` a `ThinData` as well with the trailing str instead of `SmolStr`, but that should be a followup patch to this imho.

matklad:
Yeah, the approach sounds like something we should just do!

Publishing 1.0 of thin-dst seems like a good first-step. I haven't looked at this PR closely, but one thing we want to do is to declare GreenNode like `struct GrenNode(Arc<GreenNodeData>)`. Hiding the arc inside seems good in general, and should allow to implmenet https://github.com/rust-analyzer/rust-analyzer/pull/2182/files by just changing the Cargo.toml (Ie, I expect to publish a semver-compatible version of rowan)

matklad:
BTW, I've finished most of crucial refactorings in rust-analyzer for the time being, so I'd love to prioritize this work!

CAD97:
I've got two essays due tomorrow (😨) but after that I can definitely finish this up real quick.

CAD97:
This patch cannot be semver-compatible, as `GreenElement` is no longer just an alias for `NodeOrToken` and we expose `&[GreenElement]`.

That said, I've come around to fully encapsulating the arc in `GreenNode` and should be able to make that change tomorrow.

matklad:
Oh, good catch! I should have exposed 

```
impl Iterator<Item = NodeOrToken<&GreenNode, &GreenToken>> + ExactSizeIterator
```

matklad:
As an aside, one of the lessons from iterating through a doesn of design for syntax trees is that enums are a very easy way to expose impl details in a way you can't just fix later

CAD97:
Now that we know the proper direction to take this in (and it needs to be completely refactored to do as such), I'm closing this to PR more piecewise improvement where possible. The intent is to PR the breaking changes and then make the thin-dst backend refactor entirely encapsulated.

matklad:
>The intent is to PR the breaking changes

That's an excellent idea! I'll publish it and make a PR to rust-analyzer, so that we can trivially compare perf

PR #36, opened 2019-11-14
Don't force intermediate vector allocation when constructing a node

matklad:


matklad:
Let's not merge it until we've done thin-arc shenanigans...

src/green.rs
```diff
@@ -15,9 +15,15 @@ pub struct GreenNode {
 impl GreenNode {
     /// Creates new Node.
     #[inline]
-    pub fn new(kind: SyntaxKind, children: Box<[GreenElement]>) -> GreenNode {
-        let text_len = children.iter().map(|x| x.text_len()).sum::<TextUnit>();
-        GreenNode { kind, text_len, children: children.into() }
+    pub fn new<I>(kind: SyntaxKind, children: I) -> GreenNode
```
CAD97: NB: This changes public API.

matklad:
I think this was fixed

PR #37, opened 2019-11-17
Fix SyntaxText equality

Matthias247:
This returned that Strings are equal for quite a lot of strings which
were actually not equal at all. That lead Rust Analyzer to provide
assists at times where it shouldn't have - e.g. it allowed to merge
match arms which actually did not contain the same expression.

### DIFF
--- a/src/syntax_text.rs
+++ b/src/syntax_text.rs
@@ -165,7 +165,7 @@ impl PartialEq for SyntaxText {
             return false;
         }
         let mut lhs = self.tokens_with_ranges();
-        let mut rhs = self.tokens_with_ranges();
+        let mut rhs = other.tokens_with_ranges();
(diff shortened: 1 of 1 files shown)


src/syntax_text.rs
```diff
@@ -165,7 +165,7 @@ impl PartialEq for SyntaxText {
             return false;
         }
         let mut lhs = self.tokens_with_ranges();
-        let mut rhs = self.tokens_with_ranges();
+        let mut rhs = other.tokens_with_ranges();
```
matklad: 😓

Matthias247:
> I actually remember considering writing a property test here, but thought that a handful of manual tests would do the trick.

Typically they do. But one shouldn't forget to also add some negative test cases :-)

PR #39, opened 2019-11-18
Split green into submodules

CAD97:
On top of #38. Apologies ahead of time for PR spam, but I think this is probably the best way to organize this.

Splits `green` into submodules to aid further refactoring.


matklad:
bors r+

I love optimistic PR! :)

PR #41, opened 2019-11-19
Box and cache GreenToken

CAD97:
This uses an inner `Arc` to decrease the size of `GreenToken`. This uses the same caching logic as `GreenNode` through `GreenNodeBuilder`. Do note that this means that to get a decently memory-thoughtful tree, either the use of `GreenNodeBuilder` or a manual cache in the same style is required. We should probably consider eventually exposing some version of the `Cache` implemented here, such that bottom-up builders can use it and get the deduplication for free.

The main reason for shrinking `GreenToken` is the target of shrinking `GreenElement`, as each node carries a `[GreenElement]`, and by reducing overhead per-element, we can considerably decrease overall memory usage. By boxing `GreenToken` we make room to shrink `GreenNode` and benefit from said shrinkage.

r? @matklad 

<details><summary>Size comparison</summary>

Before:

```
GreenNode    24
GreenToken   32
GreenElement 40

SyntaxNode    8
SyntaxToken   16
SyntaxElement 24
```

After:

```
GreenNode    24
GreenToken   8
GreenElement 32

SyntaxNode    8
SyntaxToken   16
SyntaxElement 24
```

### DIFF
--- a/src/green/builder.rs
+++ b/src/green/builder.rs
@@ -2,6 +2,42 @@ use crate::{cursor::SyntaxKind, NodeOrToken, SmolStr};
 
 use super::*;
 
+#[derive(Default, Debug)]
+struct Cache {
+    nodes: rustc_hash::FxHashSet<GreenNode>,
+    tokens: rustc_hash::FxHashSet<GreenToken>,
+}
+
+impl Cache {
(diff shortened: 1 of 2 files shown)


CAD97:
Next step towards a single-pointer `GreenElement` is writing up all the docs for thin-dst and publishing 1.0.0.

PR #43, opened 2019-11-19
Pack GreenElement into 1×usize PackedGreenElement

CAD97:
r? @matklad 

It is required here to tag _all_ `GreenToken` to have a low bit of `1` so that we can go from `&PackedGreenElement` to `GreenElementRef`.

### DIFF
--- a/src/green/token.rs
+++ b/src/green/token.rs
@@ -11,7 +11,7 @@ pub(crate) struct GreenTokenData {
 
 /// Leaf node in the immutable tree.
 pub struct GreenToken {
-    ptr: ptr::NonNull<GreenTokenData>,
+    pub(super) ptr: ptr::NonNull<GreenTokenData>,
 }
 
 unsafe impl Send for GreenToken {} // where GreenTokenData: Send + Sync
--- a/src/green/token.rs
+++ b/src/green/token.rs
@@ -1,41 +1,104 @@
-use std::sync::Arc;
+use std::{fmt, hash, mem::ManuallyDrop, ptr, sync::Arc};
 
 use crate::{cursor::SyntaxKind, SmolStr, TextUnit};
 
+#[repr(align(2))] // NB: this is an at-least annotation
 #[derive(Debug, PartialEq, Eq, Hash)]
-struct GreenTokenData {
+pub(crate) struct GreenTokenData {
     kind: SyntaxKind,
     text: SmolStr,
 }
 
 /// Leaf node in the immutable tree.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct GreenToken {
-    data: Arc<GreenTokenData>,
+    ptr: ptr::NonNull<GreenTokenData>,
 }
 
+unsafe impl Send for GreenToken {} // where GreenTokenData: Send + Sync
+unsafe impl Sync for GreenToken {} // where GreenTokenData: Send + Sync
+
 impl GreenToken {
+    fn add_tag(ptr: ptr::NonNull<GreenTokenData>) -> ptr::NonNull<GreenTokenData> {
+        unsafe {
+            let ptr = ((ptr.as_ptr() as usize) | 1) as *mut GreenTokenData;
+            ptr::NonNull::new_unchecked(ptr)
+        }
+    }
+
+    fn remove_tag(ptr: ptr::NonNull<GreenTokenData>) -> ptr::NonNull<GreenTokenData> {
+        unsafe {
+            let ptr = ((ptr.as_ptr() as usize) & !1) as *mut GreenTokenData;
+            ptr::NonNull::new_unchecked(ptr)
+        }
+    }
+
+    fn data(&self) -> &GreenTokenData {
+        unsafe { &*Self::remove_tag(self.ptr).as_ptr() }
+    }
+
     /// Creates new Token.
     #[inline]
     pub fn new(kind: SyntaxKind, text: SmolStr) -> GreenToken {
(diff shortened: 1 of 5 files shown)


src/green/token.rs
```diff
@@ -1,36 +1,59 @@
...
 pub struct GreenToken {
-    data: Arc<GreenTokenData>,
+    pub(super) ptr: ptr::NonNull<GreenTokenData>,
 }
 
+unsafe impl Send for GreenToken {} // where GreenTokenData: Send + Sync
+unsafe impl Sync for GreenToken {} // where GreenTokenData: Send + Sync
+
 impl GreenToken {
+    fn mangle(ptr: ptr::NonNull<GreenTokenData>) -> ptr::NonNull<GreenTokenData> {
```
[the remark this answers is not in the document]
matklad: And might be a good idea to debug-assert that the low bit is zero/one as appropriate.

matklad:
>It is required here to mangle all GreenToken to have a low bit of 1 so that we can go from &PackedGreenElement to GreenElementRef.

Oh, I see... Hm, was I wrong insisting on that we should hide `Arc` behind a wrapper? Naively, one would expect that iteration just yields pointers to nodes, without double indirection.... But I guess that just doesn't work out with Rust type system. Like, if iteration yields an `&NodeData`, you should be able to "clone" it into an `Arc<NodeData>` (using shared from this trick), but that doesn't work out type-wise, because clone can only remove `&`.


I think the current approach is the best one to start with, but maybe we should think about making it more straightforward on the machine level (no double pointers), without forking the types on the Rust level. What bites us is that `Clone` works for `&T -> T` and `T -> T`, but we want to `&T -> U` and `U -> U`, which is a `ToOwned`, `Clone` pair, which is more awkward to use

matklad:
Hm, OTOH, these are green nodes, so perhaps we don't have to pay too much attention to the Rust-level API

CAD97:
@matklad yeah, that's the foot: either you expose `String`/`str` separately, or everyone has to deal in `&String`. Note that `cursor::NodeData` doesn't have to point at `GreenNode(Arc<Inner>)`, though; it can point directly to the inner data. So the indirection only exists when manipulating green nodes directly.

After we get this encapsulated version pushed through, I've an itch to scratch basically rebuilding rowan in an space-optimized-from-first-principles manner, so it'll be interesting to compare that experiment once it's done.

src/green/token.rs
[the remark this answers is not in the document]
CAD97: FWIW, calling `add_tag` a second time is a NOP. The `#[repr(align(2))]` makes sure the low bit is for alignment.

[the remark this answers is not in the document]
CAD97: FWIW, calling `add_tag` a second time is a NOP. The `#[repr(align(2))]` makes sure the low bit is for alignment.

matklad:
>After we get this encapsulated version pushed through, I've an itch to scratch basically rebuilding rowan in an space-optimized-from-first-principles manner, so it'll be interesting to compare that experiment once it's done.

Yeah, I feel like, with thread-local SyntaxNode, and pointer-wide interned green elements, rowan is mostly done, design wise, so it makes sense to polish this approach, and maybe publish 1.0. The only problem is that I've though exactly that before :-)

One potential improvement we still can make is adding ability to parse things lazily. Basically, for things like functions, we can store bodies literally as `String`, and parse them into propper green nodes on the first access.

src/green/token.rs
```diff
@@ -1,36 +1,59 @@
...
 
 /// Leaf node in the immutable tree.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct GreenToken {
-    data: Arc<GreenTokenData>,
+    pub(super) ptr: ptr::NonNull<GreenTokenData>,
 }
 
+unsafe impl Send for GreenToken {} // where GreenTokenData: Send + Sync
+unsafe impl Sync for GreenToken {} // where GreenTokenData: Send + Sync
```
matklad: Can't we just add this where bound?

src/green/token.rs
```diff
@@ -1,36 +1,59 @@
...
+
 impl GreenToken {
+    fn add_tag(ptr: ptr::NonNull<GreenTokenData>) -> ptr::NonNull<GreenTokenData> {
+        unsafe {
```
matklad: let's add debug asserts

matklad: And maybe use `^ 1` as a convenient way to flip a bit

src/green/token.rs
CAD97: No, `GreenToken` is public and `GreenTokenData` is private.

CAD97:
Dropping this here since I don't really have a better place to put it, but I had an idea for

> One potential improvement we still can make is adding ability to parse things lazily. Basically, for things like functions, we can store bodies literally as String, and parse them into propper green nodes on the first access.

Adding another "type" of node to the tree is problematic because everything currently assumes two types: leaf (token) or non-leaf (node). So.... just make a delayed blob a token!

There are a couple ways to handle this:

- Reserve a single kind for `SyntaxKind::Blob`. This is semantically allowed to replace any kind of node in the tree and represents that it isn't parsed yet.
- Lower the "kind space" to u15. If the high bit of the u16 kind is set, this is an unparsed node/token of the kind indicated in the remaining 15 bits.
- Distinguish between "token" kinds and "node" kinds. If a token has a "node" kind, this is an unparsed node.

I'm going to try out the third approach some in my experimental rewrite.

PR #44, opened 2019-11-25
First principles

CAD97:


src/helpers/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    pub fn is_token(&self) -> bool {
+        self.as_token().is_some()
+    }
+
+    pub fn as_ref(&self) -> NodeOrToken<&Node, &Token> {
+        match *self {
+            NodeOrToken::Node(ref node) => NodeOrToken::Node(node),
+            NodeOrToken::Token(ref token) => NodeOrToken::Token(token),
```
matklad: I think you can remove `*` and `ref` to get the same effect with just match ergonomics

src/helpers/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+        n: impl FnOnce(Node) -> N,
+        t: impl FnOnce(Token) -> T,
+    ) -> NodeOrToken<N, T> {
+        match self {
+            NodeOrToken::Node(node) => NodeOrToken::Node(n(node)),
+            NodeOrToken::Token(token) => NodeOrToken::Token(t(token)),
+        }
+    }
+
+    pub fn deref(&self) -> NodeOrToken<&Node::Target, &Token::Target>
```
matklad: as_deref: https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref

src/helpers/mod.rs
CAD97: I specifically chose to use `ref` here to make it more obvious, as the match ergonomics version looks like a no-op:

```rust
        match self {
            NodeOrToken::Node(node) => NodeOrToken::Node(node),
            NodeOrToken::Token(token) => NodeOrToken::Token(token),
        }
```

PR #46, opened 2019-12-01
Fix SyntaxText equality for substring

CAD97:
I found another wrong case in `SyntaxText`'s equality routine, with `SyntaxText == &str`. This case wasn't being tested.


matklad:
bors r+

:man_facepalming:

PR #47, opened 2019-12-02
Formatting

kjeremy:
cargo fmt


matklad:
we should add this to CI config

CAD97:
@matklad if you don't get to it first I can set it up on GitHub Actions later this week.

kjeremy:
@CAD97 
Latest stable 1.4.8-stable

PR #51, opened 2020-01-17
take empty tokens into account

vemoo:
This is usefull if you are creating empty tokens to represent virtual tokens that were added to make the ast make sense.

Before, for the tokens `["A", "+", ""]` calling `token_at_offset(2)` would return `TokenAtOffset::Single("+")``.
Now it returns `TokenAtOffset::Between("+", "")`.

I've added a simple test, and I've also run the `rust-analyzer` tests with the modified `rowan` crate.

Something to take into account with this implementation is that if `token_at_offset` is called in a non root node with `offset` at the start or end of the node, `TokenAtOffset::Between` will be returned with a token from outside this node.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -476,14 +476,15 @@ impl SyntaxNode {
                         }
                     }
                     SyntaxElement::Node(node) => {
-                        if node.text_range().contains_inclusive(offset) {
+                        let range = node.text_range();
+                        if !range.is_empty() && range.contains_inclusive(offset) {
                             container_element = node;
                             continue 'outer;
                         }
                     }
                 }
             }
-            unreachable!()
+            unreachable!();
         }
     }
 
@@ -825,6 +826,25 @@ mod tests {
         builder.token(SyntaxKind(7), SmolStr::from(text))
     }
 
+    fn check(actual: TokenAtOffset<SyntaxToken>, expected: TokenAtOffset<&str>) {
+        match (actual, expected) {
+            (
+                TokenAtOffset::Between(actual_left, actual_right),
+                TokenAtOffset::Between(expected_left, expected_right),
+            ) => {
+                assert_eq!(actual_left.text(), expected_left);
+                assert_eq!(actual_right.text(), expected_right);
+            }
+            (TokenAtOffset::Single(actual), TokenAtOffset::Single(expected)) => {
+                assert_eq!(actual.text(), expected);
+            }
+            (TokenAtOffset::None, TokenAtOffset::None) => {}
+            (actual, _) => {
+                assert!(false, "unexpected {:#?}", actual);
+            }
+        }
+    }
+
     #[test]
     fn token_at_offset_with_empty_token() {
         let mut builder = GreenNodeBuilder::new();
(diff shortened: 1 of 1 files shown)


src/cursor.rs
```diff
@@ -456,25 +453,37 @@ impl SyntaxNode {
...
+                            container_element = node;
+                            continue 'outer;
+                        }
+                    }
                 }
-                _ => unreachable!(),
             }
-        } else {
-            left.token_at_offset(offset)
+            unreachable!()
```
jD91mZM2: This should theoretically be able to be reached if there's a node with zero children... Right?

jD91mZM2 (changes requested):
I like it! I don't know exactly about the pros/cons of supporting empty tokens, I'll let @matklad decide on that :)

src/cursor.rs
vemoo: for nodes without children that shouldn't be reached because of https://github.com/rust-analyzer/rowan/blob/be215ee19c8818e854164580f34ad3ef0d78e75a/src/cursor.rs#L452-L454

matklad:
I am very hesitant about supporting empty tokens, as they open all kinds of nasty edge cases. 

In particular, with empty tokens, `TokenAtOffset` type does not really make sense, as there may be arbitrary many empty tokens at a given position, and, if we fully support empty tokens, we should return a `Vec` or an `Iterator` here.

Initially (way back in `fall`) I've used empty tokens and nodes extensively (for example, syntax errors like "missing `;`" were modeled as empty nodes), but that caused a lot of problems down the line, exectly because the API became non-intuitive due to all of edge cases. 

I'd love to go further and even forbid empty internal nodes, but that unfortunately hits a really horrible edge-case: an empty file. I am even wondering if the API would be better if we had an explicit EMPTY_FILE token for this case....

vemoo:
But returning `TokenAtOffset::Single` in the presence of empty tokens also doesn't make sense.
For the example of tokens `["A", "+", "", "*", "C"]` `token_at_offset(2)` before this change returns `TokenAtOffset::Single("+")` but why not `TokenAtOffset::Between("+", "*")` or `TokenAtOffset::Single("*")`?

I realise there could be arbitrary many empty tokens, but `TokenAtOffset::Between("+", "")` could be considered left biased in a way. With empty tokens the return type will always be somthing like `TokenAtOffset::Between(<non_empty>, <empty>)`.

Other alternative would be to disallow them when building them.

src/cursor.rs
vemoo: I did find another issue related to empty nodes. I fixed it and added a couple of tests. https://github.com/rust-analyzer/rowan/pull/51/commits/2bfdd9ec4cf2c7882c6ba7ffb65425aa9c978fab

matklad:
>Other alternative would be to disallow them when building them.

Yeah, we currently "soft-disalow" them, in that code assumes that tokens are non-empty. We don't have any active assetions though, because nothing actively *breaks* with empty tokens. If one has to have empty tokens, one can write custom algorithms which do something resonable

CAD97:
When I first learned indexes, I learned them like the cursor of a text editor:

```
|a|b|c|d|e|f|g|
^ ^ ^ ^ ^ ^ ^ ^
0 1 2 3 4 5 6 7
```

Per this understanding of the offset (cursor position), then `TokenAtOffset::Between` is definitely correct for when the cursor is next to a zero-width token.

For example, given the tokens `["a", "", "b"]` and offset 1, the cursor could either be at `a|␣b` or `a␣|b`. So for this, `TokenAtOffset::Between("a", "")` or `::Between("", "a")` would be correct, you could argue `::Between("a", "b")` (but I'd argue against), and `::Single("a")` is just incorrect (the position is not within the span of the `a` token, but after it).

jD91mZM2 (approved):

BenjaminBrienen:
is this still valid?

PR #52, opened 2020-02-22
Maybe a mistake fix

Veetaha:
?

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -447,7 +447,7 @@ impl SyntaxNode {
         // recursion with a loop.
         let range = self.text_range();
         assert!(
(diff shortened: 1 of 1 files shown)


CAD97:
I don't think this is wrong. When we receive an offset equivalent to the length of the selected span, the correct token to say is present there is `None`, as is handled for the empty span case directly below the assert.

An actual example (other than an empty span) that passes the assert on master and is caught by this patch would help in determining the correct/expected behavior here.

Veetaha:
Oh, I just read the precondition comment and saw the assertion that kinda conflict with it...

matklad:
Yep, text offset can point right at the end of the node/file, it can be equal to the length of the thing.

PR #54, opened 2020-04-14
Update text_unit to text-size

CAD97:


src/green/element.rs
```diff
@@ -63,11 +63,12 @@ impl GreenElement {
     pub fn kind(&self) -> SyntaxKind {
         self.as_ref().kind()
     }
+}
 
-    /// Returns length of the text covered by this element.
+impl TextLen for &'_ GreenElement {
     #[inline]
-    pub fn text_len(&self) -> TextUnit {
```
matklad, line 69: This should be kept as a public inherent method and used instead of `TextUnit::of(&node)`. Otherwise, r=me

matklad: Oh, let's allso change the version of the crate itself to `0.next.0-pre`, so that I can publish it?

CAD97: Did so.

PR #56, opened 2020-04-26
Rename serde feature flag to `serde`

CAD97:
The `serde1` feature flag should be removed with the next version bump.

(I'm working on a proof of implementing rowan's green tree by wrapping sorbus's green tree, so that will probably be the next version bump.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,11 +11,18 @@ edition = "2018"
 rustc-hash = "1.0.1"
 text-size = "1.0.0"
 smol_str = "0.1.10"
-serde = { version = "1.0.89", optional = true, default-features = false }
 thin-dst = "1.0.0"
+
+[dependencies.serde_] # renamed so feature flag can be "serde"
+package = "serde"
+version = "1.0.89"
+optional = true
+default-features = false
 
 [dev-dependencies]
 m_lexer = "0.0.4"
 
 [features]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -11,11 +11,18 @@ edition = "2018"
...
+package = "serde"
+version = "1.0.89"
+optional = true
+default-features = false
+
 [dev-dependencies]
 m_lexer = "0.0.4"
 
 [features]
-serde1 = [ "serde", "text-size/serde" ]
```
matklad, line 21: I named the feature `serde1` because, once, there was and idea that this'll work better with future `serde2`... Not sure what's the best practice here really

CAD97: The standard recommendation seems to be to [always just call the feature flag serde](https://github.com/rust-lang/api-guidelines/issues/180).

CAD97: See also the API guideline [C-feature](https://rust-lang.github.io/api-guidelines/naming.html#c-feature) which I recall mentioning preferring `serde` to `serde1` previously as well.

matklad:
Looking at this more, I think we shouldn't do this. 

* In the grand scheme of things, this doesn't matter
* I am not a huge fan of work-around personally. Between a nice API with awkward implementation and somewhat awkward API with nice impl, I prefer the second case for all but the most re-usable crates.

PR #62, opened 2020-06-18
migrate from thin-dst to slice-dst and erasable

CAD97:
thin-dst is deprecated in favor of using slice-dst and erasable, which provide a more principled and build-your-abstraction API.

This just ports the existing structure to use `erasable::Thin<Arc<slice_dst::SliceWithHeader<_, _>>>` instead of `thin_dst::ThinArc<_, _>`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,11 +8,12 @@ description = "Library for generic lossless syntax trees"
 edition = "2018"
 
 [dependencies]
+erasable = "1.2.1"
 rustc-hash = "1.0.1"
-text-size = "1.0.0"
-smol_str = "0.1.10"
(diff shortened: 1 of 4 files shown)


CAD97:
bors: r+

This is a minor refactoring and a pre-requisite of exposing a GC for the cache.

CAD97:
CI passed, but the .org Travis seems to have not been able to report that back. https://travis-ci.org/github/rust-analyzer/rowan/builds/699851971

cc @matklad we'll probably need to either migrate to .com Travis or GHA.

Merging manually.
bors: cancel

CAD97:
bors: r-

PR #63, opened 2020-06-19
Add ability to gc node caches

### LINKED ISSUE #53: Exporting some version of Cache
In PR #41 it was mentioned

> We should probably consider eventually exposing some version of the Cache implemented here, such that bottom-up builders can use it and get the deduplication for free.

This would be seem very helpful in trying to integrate lalrpop.

CAD97:
Originally implemented in sorbus here: https://github.com/CAD97/sorbus/pull/46

Also exposes the `NodeCache` (closes #53) so that the gc is usable. While `GreenNodeBuilder` has had `with_cache` for a while, it's been impossible to _actually_ share the cache, as there was no way to get a cache to share!

### DIFF
--- a/src/green/builder.rs
+++ b/src/green/builder.rs
@@ -1,4 +1,4 @@
-use rustc_hash::FxHashSet;
+use hashbrown::HashMap;
 
 use crate::{
     green::{GreenElement, GreenNode, GreenToken, SyntaxKind},
@@ -7,8 +7,8 @@ use crate::{
 
 #[derive(Default, Debug)]
 pub struct NodeCache {
-    nodes: FxHashSet<GreenNode>,
-    tokens: FxHashSet<GreenToken>,
+    nodes: HashMap<GreenNode, ()>,
+    tokens: HashMap<GreenToken, ()>,
 }
 
 impl NodeCache {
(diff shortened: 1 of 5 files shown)


src/green/builder.rs
```diff
@@ -7,8 +7,8 @@ use crate::{
 
 #[derive(Default, Debug)]
 pub struct NodeCache {
-    nodes: FxHashSet<GreenNode>,
-    tokens: FxHashSet<GreenToken>,
+    nodes: HashMap<GreenNode, ()>,
+    tokens: HashMap<GreenToken, ()>,
```
kjeremy, line 11: Are these `HashMap` just because of `drain_filter`?

CAD97: Yes (unfortunately).

A further optimization step in the pipeline will take further advantage of `HashMap::raw_entry_mut`, so the switch to using set-like maps is probably inevitable, though.

simonvandel:
What is the effect on peak memory usage for analysis stats using rust analyzer?

CAD97:
It's not going to have an immediate impact because r-a doesn't (IIUC) yet hold on to or reuse node caches.

Additionally, it will only potentially have an impact when holding onto a node cache after the deletion of a parse tree. We aren't really generating any garbage to clean up; it's only useful if the cache outlives the tree(s) that it's built.

BenjaminBrienen:
is this still valid?

PR #65, opened 2020-06-26
Add cargo fmt to travis build config

Atul9:
## Note: To fix the cargo fmt build error the https://github.com/rust-analyzer/rowan/pull/64 PR needs to be merged first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,11 @@ rust:
   - beta
 
 script:
+  - |
+    if [[ $TRAVIS_RUST_VERSION == "stable" ]]
+    then
+      rustup component add rustfmt
+      cargo fmt -- --check
+    fi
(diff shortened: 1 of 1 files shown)


matklad:
Let's add this as a test instead, I am not a fan of encoding stuff into the CI config. See rust-analyzer as an example:

https://github.com/rust-analyzer/rust-analyzer/blob/7488cd6a1b3604d6b4f11a7f83b6006a34cda0c0/xtask/tests/tidy.rs#L33-L38

Atul9:
ok, I am checking ^ rust-analyzer `run_rustfmt` & other functions which are called. In which file should it be added in this project?

Atul9:
@matklad ok, I am checking ^ rust-analyzer run_rustfmt & other functions which are called.    
In which file should it be added in this project?

PR #66, opened 2020-07-02
Move from TravisCI to GHA

CAD97:


matklad:
LGTM!

CAD97:
I... don't know why the action isn't triggering though... 😅

CAD97:
It ran and passed when I pushed to master in my fork: https://github.com/cad97/rowan/actions/runs/155876775

So merging this as green, hopefully that will let GH pick up that the workflow exists. (If not, there's probably a setting you'll need to change.)

CAD97:
Actions are running fine now that the workflow is on the target branch: https://github.com/rust-analyzer/rowan/actions/runs/155877867

bjorn3:
Github actions isn't activated untill you push a workflow to the default branch.

PR #67, opened 2020-07-02
Add bors.toml

CAD97:
I have no idea how bors ever worked on this repository when I don't think this repository has ever had a bors.toml.


CAD97:
bors: r+

PR #68, opened 2020-07-02
Add fmt tidy test

CAD97:
Supersedes/closes #65 by implementing this as a test rather than in the CI configuration. Impl taken from r-a but dumbed down to the absolute minimum.

r? @matklad

### DIFF
--- a/tests/tidy.rs
+++ b/tests/tidy.rs
@@ -1,0 +1,46 @@
+use std::{
+    env,
+    path::{Path, PathBuf},
+    process::{Command, Stdio},
+};
+
+fn project_root() -> PathBuf {
+    PathBuf::from(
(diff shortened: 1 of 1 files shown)


tests/tidy.rs
```diff
@@ -0,0 +1,46 @@
...
+    process::{Command, Stdio},
+};
+
+fn project_root() -> PathBuf {
+    PathBuf::from(
+        env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| env!("CARGO_MANIFEST_DIR").to_owned()),
+    )
+}
+
+fn run(cmd: &str, dir: impl AsRef<Path>) -> Result<(), ()> {
```
matklad, line 13: For non-public APIs, I prefer to avoid `AsRef` polymorphism, to save on compile times. But this probably doesn't matter. 

bors r+

PR #100, opened 2021-03-28
Optimised TokenText (fat ptr)

iDawer:
Following https://github.com/rust-analyzer/rust-analyzer/pull/8209
Basically it is a fat pointer `Nonnull<str>` with owning semantics, and uses field offset to get rid of indirection.
The question is: It's not that mad, is it? I need someone to agree with me so I could have finished the implementation. :smile:

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -395,6 +395,26 @@ impl<H, T> ThinArc<H, T> {
 
         ThinArc { ptr: unsafe { ptr::NonNull::new_unchecked(ptr) }, phantom: PhantomData }
     }
+
+    /// ### Safety:
+    ///  - `[T]` must have no interrior mutability.
+    ///  - slice ptr must come from a leaked ThinArc.
+    pub(crate) unsafe fn from_raw_slice(raw: *mut [T]) -> Self {
+        let inner_to_data_offset = offset_of!(ArcInner<HeaderSlice<H, [T; 0]>>, data);
+        let data_to_slice_offset = offset_of!(HeaderSlice<H, [T; 0]>, slice);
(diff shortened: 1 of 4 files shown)


matklad:
Yeah, I think this belongs here! Although let's start with a double-indirection safe implementation firs. I won't expect this to be a bottleneck, and I would expect a lot of churn in rowan. 

But I do think we should have something along this lines long-term. We can even do a crazier thing, and *borrow* nodes of immutable syntax trees. Hm, now that I've spelled this out, I wonder if we actually should just return `Cow<str, String>` from that rust-analyzer method? This probably requires exposing `is_mutable` getter...

iDawer:
I'll open a new PR with with the rust-analyzer's implementation

iDawer:
> I'll open a new PR with with the rust-analyzer's implementation

nvm

> We can even do a crazier thing, and borrow nodes of immutable syntax trees. Hm, now that I've spelled this out, I wonder if we actually should just return Cow<str, String> from that rust-analyzer method?

Borrowing nodes of immutable syntax trees means `NodeData.mutable` once created never changes, And mixing immutable syntax nodes with mutable ones is not allowed. Are these invariants hold in code?
This really intrigued me. I would investigate on it before moving ra's TokenText.

matklad:
>NodeData.mutable once created never changes, And mixing immutable syntax nodes with mutable ones is not allowed.

Yup, the relevant place to document that would be: https://github.com/rust-analyzer/rowan/blob/clone-for-update/src/cursor.rs#L17-L75.

Ohhhhhh. Oh wow, we can actually change the `green` method in rowan itself to return a `Cow<&GreenNode, GreenNode>`. Sweeeeet!

PR #101, opened 2021-04-01
Borrow nodes of immutable syntax trees

iDawer:
Borrow green nodes of immutable syntax trees.
It requires some clarified invariants inside the `cursor` module. But overall, I think, it is safe for an API consumer.

(I couldn't find the latest 0.13.0-pre.3 version commit. Where should this go?)

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -73,8 +73,16 @@
 // tokens the underlying green token actually is immutable, so we can, and do
 // return `&str`.
 //
+// Invariants [must not leak outside of the module]:
+//    - Mutability is the property of the whole tree. Intermixing elements that
+//      differ in mutability is not allowed.
+//    - Mutability property is persistent.
+//    - References to the green elements' data are not exposed into public API
+//      when the tree is mutable.
+//    - TBD
 
 use std::{
+    borrow::Cow,
     cell::Cell,
     fmt,
     hash::{Hash, Hasher},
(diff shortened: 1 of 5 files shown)


lnicola:
Yeah, `0.13.0-pre.3` was already published, you might want to go to `-pre.4` and rebase later: https://crates.io/crates/rowan/versions

lnicola:
You can probably rebase now.

matklad:
LGTM!

iDawer:
Should I bump version?

src/cursor.rs
```diff
@@ -574,8 +583,12 @@ impl SyntaxNode {
     }
 
     #[inline]
-    pub fn green(&self) -> GreenNode {
-        self.green_ref().to_owned()
+    pub fn green(&self) -> Cow<'_, GreenNodeData> {
+        let green_ref = self.green_ref();
```
iDawer, line 587: `SyntaxNode::green_ref` should be marked unsafe, otherwise this looks too suspicious. I think it can be done later.

matklad: Probably! I kind-of threw the towel on managing "internal" unsafety boundary here, but marking that one `unsafe` surely seems good!

matklad:
merged into master and publishing as pre.4

iDawer:
Whoops, seems like this PR recreated the clone-for-update branch, and now the published version does not contain the changes

matklad:
heh, @matklad: 0, git: 1. 

master should be better now

PR #105, opened 2021-06-30
Bump hashbrown

kjeremy:
This should help compile times and allow us to use the newer version in rust-analyzer


lnicola:
Might have been worth bumping the patch version too.

kjeremy:
Oh yes I should have done that so we could pull it in to rust analyzer.
I'll send another PR out later today if someone doesn't beat me to it.

lnicola:
Filed  #106.

Issue #108, opened 2021-07-29
Miri UB with -Zmiri-track-raw-pointers


pandaman64:
Hi! I got UB when running Miri with raw pointer tracking enabled (`-Zmiri-track-raw-pointers`) against the invocation of `GreenNodeBuilder::token`. The UB is not reported when running Miri without raw pointer tracking.
To reproduce the UB, run the following function with `MIRIFLAGS='-Zmiri-track-raw-pointers' cargo +nightly miri run`.

```rust
use rowan::{GreenNodeBuilder, SyntaxKind};

fn main() {
    let mut builder = GreenNodeBuilder::new();
    builder.start_node(SyntaxKind(0));
    builder.token(SyntaxKind(1), "foo");
}
```

Here is the backtrace:
```rust
error: Undefined Behavior: no item granting write access to tag <3629> at alloc1681+0x18 found in borrow stack.
   --> /home/pan/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:887:9
    |
887 |         copy_nonoverlapping(&src as *const T, dst, 1);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item granting write access to tag <3629> at alloc1681+0x18 found in borrow stack.
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
            
    = note: inside `std::ptr::write::<u8>` at /home/pan/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:887:9
    = note: inside `rowan::arc::ThinArc::<rowan::green::token::GreenTokenHead, u8>::from_header_and_iter::<std::str::Bytes>` at /home/pan/.cargo/registry/src/github.com-1ecc6299db9ec823/rowan-0.12.6/src/arc.rs:382:21
    = note: inside `rowan::GreenToken::new` at /home/pan/.cargo/registry/src/github.com-1ecc6299db9ec823/rowan-0.12.6/src/green/token.rs:110:19
[...]

PR #109, opened 2021-07-29
Use some ptr::addr_of_mut for taking ptr addr

CAD97:
Relevant to #108, but I'm not certain if it addresses the miri failure or not. (Can't check because of https://github.com/rust-lang/miri/issues/705.)

Either way, this is a positive change, as previously we were creating `&mut` to uninitialized memory locations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.13.0-pre.7"
+version = "0.13.0-pre.8"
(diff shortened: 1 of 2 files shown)


pandaman64:
I confirmed that this patch fixes #108. Thanks!

lnicola:
Can you update the version in `Cargo.toml`?

CAD97:
Version bumped

lnicola:
And it's published, thanks!

PR #111, opened 2021-08-23
Fix NodeCache which reduced caching efficiency, while making it faster

matklad:
Turns out that we accidentally use two different hash functions in
NodeCache, which is wrong in a fun way. This was discovered while trying
to optimize equality to be based on pointer comparisons.

See https://youtu.be/bVoHUzswKog for details

bors r+


CAD97:
Whoops! :smile:

matklad:
Indeed!

PR #112, opened 2021-08-23
Fix memory leak in NodeData constructor

Veykril:
Drop SyntaxNode in mutable trees when reusing a NodeData


matklad:
Oh wow, I thought that my caching bug was bad, but this seems worse, awesome job @Veykril

Veykril:
This was hiding pretty well, took quite some time to find.

PR #114, opened 2021-09-19
Return `PreorderWithTokens` iterator in `SyntaxNode::preorder_with_tokens`

Veykril:
To allow skipping subtrees


CAD97:
Seems reasonable to me.

bors: r+
(if I have the permissions, I don't recall)

PR #119, opened 2021-10-03
Reduce the number of allocations needed to find a specific child/sibling

theo-lw:
A very common operation in rust-analyzer is finding a child/sibling of a `SyntaxNode` that matches a specific `SyntaxKind`. This is currently implemented by iterating through children and finding the first matching node. However, this causes unnecessary allocations - the children that don't match are allocated and quickly discarded. 

This PR aims to reduce these unnecessary allocations by providing methods to iterate through only the children/siblings that match the desired `SyntaxKind`. Children/siblings that don't match the desired `SyntaxKind` are not allocated in these iterators.

So, I've only implemented this optimization for iterating through children, but this optimization already seems quite promising. Below is the DHAT output from running rust-analyzer's integrated completion benchmarks.

DHAT output before this change (on the [master branch](https://github.com/theo-lw/rust-analyzer/tree/master) in my fork):

```
==196375== Total:     15,179,316,140 bytes in 107,064,174 blocks
==196375== At t-gmax: 587,308,085 bytes in 2,611,269 blocks
==196375== At t-end:  845,424 bytes in 1,480 blocks
==196375== Reads:     33,131,042,097 bytes
==196375== Writes:    17,949,507,478 bytes
```

DHAT output after this change (on this [branch](https://github.com/theo-lw/rust-analyzer/tree/optimize-child-iteration) where I used `children_matching` in `support::child`).

```
==82931== Total:     14,585,180,134 bytes in 97,752,267 blocks
==82931== At t-gmax: 585,555,693 bytes in 2,599,462 blocks
==82931== At t-end:  845,584 bytes in 1,483 blocks
==82931== Reads:     32,637,529,753 bytes
==82931== Writes:    17,171,029,456 bytes
```

This is roughly a 8.5% decrease in total blocks allocated and a 4% decrease in the total bytes allocated.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -389,24 +389,30 @@ impl<L: Language> Iterator for SyntaxNodeChildren<L> {
 }
 
 impl<L: Language> SyntaxNodeChildren<L> {
-    pub fn by_kind<F: Clone + Fn(SyntaxKind) -> bool>(
+    pub fn by_kind<'a>(
         self,
-        matcher: F,
-    ) -> SyntaxNodeChildrenByKind<F, L> {
+        matcher: &'a impl Fn(SyntaxKind) -> bool,
+    ) -> SyntaxNodeChildrenByKind<'a, L> {
         SyntaxNodeChildrenByKind { raw: self.raw.by_kind(matcher), _p: PhantomData }
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct SyntaxNodeChildrenByKind<F: Clone + Fn(SyntaxKind) -> bool, L: Language> {
-    raw: cursor::SyntaxNodeChildrenByKind<F>,
+#[derive(Clone)]
+pub struct SyntaxNodeChildrenByKind<'a, L: Language> {
+    raw: cursor::SyntaxNodeChildrenByKind<&'a dyn Fn(SyntaxKind) -> bool>,
     _p: PhantomData<L>,
 }
 
-impl<F: Clone + Fn(SyntaxKind) -> bool, L: Language> Iterator for SyntaxNodeChildrenByKind<F, L> {
+impl<'a, L: Language> Iterator for SyntaxNodeChildrenByKind<'a, L> {
     type Item = SyntaxNode<L>;
     fn next(&mut self) -> Option<Self::Item> {
         self.raw.next().map(SyntaxNode::from)
+    }
+}
+
+impl<'a, L: Language> fmt::Debug for SyntaxNodeChildrenByKind<'a, L> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SyntaxNodeChildrenByKind").finish()
     }
 }
 
@@ -424,26 +430,30 @@ impl<L: Language> Iterator for SyntaxElementChildren<L> {
 }
 
 impl<L: Language> SyntaxElementChildren<L> {
-    pub fn by_kind<F: Clone + Fn(SyntaxKind) -> bool>(
+    pub fn by_kind<'a>(
         self,
-        matcher: F,
-    ) -> SyntaxElementChildrenByKind<F, L> {
+        matcher: &'a impl Fn(SyntaxKind) -> bool,
+    ) -> SyntaxElementChildrenByKind<'a, L> {
         SyntaxElementChildrenByKind { raw: self.raw.by_kind(matcher), _p: PhantomData }
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct SyntaxElementChildrenByKind<F: Clone + Fn(SyntaxKind) -> bool, L: Language> {
-    raw: cursor::SyntaxElementChildrenByKind<F>,
+#[derive(Clone)]
+pub struct SyntaxElementChildrenByKind<'a, L: Language> {
+    raw: cursor::SyntaxElementChildrenByKind<&'a dyn Fn(SyntaxKind) -> bool>,
     _p: PhantomData<L>,
 }
 
-impl<F: Clone + Fn(SyntaxKind) -> bool, L: Language> Iterator
-    for SyntaxElementChildrenByKind<F, L>
-{
+impl<'a, L: Language> Iterator for SyntaxElementChildrenByKind<'a, L> {
     type Item = SyntaxElement<L>;
     fn next(&mut self) -> Option<Self::Item> {
         self.raw.next().map(NodeOrToken::from)
+    }
+}
+
+impl<'a, L: Language> fmt::Debug for SyntaxElementChildrenByKind<'a, L> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SyntaxElementChildrenByKind").finish()
     }
 }
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -391,7 +391,7 @@ impl<L: Language> Iterator for SyntaxNodeChildren<L> {
 impl<L: Language> SyntaxNodeChildren<L> {
     pub fn by_kind<'a>(
         self,
-        matcher: &'a impl Fn(SyntaxKind) -> bool,
+        matcher: &'a dyn Fn(SyntaxKind) -> bool,
     ) -> SyntaxNodeChildrenByKind<'a, L> {
         SyntaxNodeChildrenByKind { raw: self.raw.by_kind(matcher), _p: PhantomData }
     }
@@ -432,7 +432,7 @@ impl<L: Language> Iterator for SyntaxElementChildren<L> {
 impl<L: Language> SyntaxElementChildren<L> {
     pub fn by_kind<'a>(
         self,
-        matcher: &'a impl Fn(SyntaxKind) -> bool,
+        matcher: &'a dyn Fn(SyntaxKind) -> bool,
     ) -> SyntaxElementChildrenByKind<'a, L> {
         SyntaxElementChildrenByKind { raw: self.raw.by_kind(matcher), _p: PhantomData }
     }
--- a/src/api.rs
+++ b/src/api.rs
@@ -392,18 +392,18 @@ impl<L: Language> SyntaxNodeChildren<L> {
     pub fn by_kind<F: Clone + Fn(SyntaxKind) -> bool>(
         self,
         matcher: F,
-    ) -> SyntaxNodeChildrenMatching<F, L> {
-        SyntaxNodeChildrenMatching { raw: self.raw.by_kind(matcher), _p: PhantomData }
+    ) -> SyntaxNodeChildrenByKind<F, L> {
+        SyntaxNodeChildrenByKind { raw: self.raw.by_kind(matcher), _p: PhantomData }
     }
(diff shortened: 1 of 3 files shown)


theo-lw:
DHAT output of the integrated completion benchmarks with the latest commits:

```
==73109== Total:     13,722,837,242 bytes in 84,207,767 blocks
==73109== At t-gmax: 585,402,987 bytes in 2,596,664 blocks
==73109== At t-end:  846,800 bytes in 1,487 blocks
==73109== Reads:     31,817,411,560 bytes
==73109== Writes:    16,038,960,745 bytes
```

Roughly a 21% decrease in total blocks allocated, a 9.6% decrease in total bytes allocated.

src/api.rs
```diff
@@ -94,7 +94,7 @@ impl<L: Language> From<SyntaxToken<L>> for SyntaxElement<L> {
     }
 }
 
-impl<L: Language> SyntaxNode<L> {
+impl<'a, L: Language> SyntaxNode<L> {
```
CAD97, line 97: Shouldn't this `'a` be introduced at `pub fn children_with_tokens_matching`? I'm surprised that an unused lifetime like this is allowed.

CAD97 (approved):
> Roughly a 21% decrease in total blocks allocated, a 9.6% decrease in total bytes allocated.

👀 that's pretty significant!

Needs a `cargo fmt`, but otherwise LGTM.

It might also be interesting to instead expose this as e.g. `.children().filter_by_kind(|kind| ..)` rather than `.children_matching(|kind| ..)`, but I don't really have a strong preference here. 

r? @matklad?

src/api.rs
theo-lw: It should. Fixed!

src/api.rs
```diff
@@ -388,6 +388,34 @@ impl<L: Language> Iterator for SyntaxNodeChildren<L> {
     }
 }
 
+impl<L: Language> SyntaxNodeChildren<L> {
+    pub fn by_kind<'a>(
+        self,
+        matcher: &'a impl Fn(SyntaxKind) -> bool,
```
Veykril, line 394:
```suggestion
        matcher: &'a dyn Fn(SyntaxKind) -> bool,
```
since this is internally stored as a trait object already we don't gain anything from this being generic right?

src/api.rs
```diff
@@ -401,6 +429,34 @@ impl<L: Language> Iterator for SyntaxElementChildren<L> {
     }
 }
 
+impl<L: Language> SyntaxElementChildren<L> {
+    pub fn by_kind<'a>(
+        self,
+        matcher: &'a impl Fn(SyntaxKind) -> bool,
```
Veykril, line 435:
```suggestion
        matcher: &'a dyn Fn(SyntaxKind) -> bool,
```

src/cursor.rs
```diff
@@ -388,6 +388,24 @@ impl NodeData {
             })
         })
     }
+
+    fn next_sibling_by_kind(&self, matcher: &impl Fn(SyntaxKind) -> bool) -> Option<SyntaxNode> {
+        let mut siblings = self.green_siblings().enumerate();
+        let index = self.index() as usize;
+
+        siblings.nth(index);
```
Veykril, line 396: nit but using `.skip(index + 1).find_map(...` might be clearer

src/cursor.rs
```diff
@@ -1133,46 +1243,130 @@ impl From<SyntaxToken> for SyntaxElement {
...
+            next
+        })
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct SyntaxElementChildren {
+    parent: SyntaxNode,
     next: Option<SyntaxElement>,
+    next_initialized: bool,
```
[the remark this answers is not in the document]
Veykril: Sounds odd to me that using `first_child_or_token_by_kind` with these extra changes is so much faster than seeking to the next sibling via `next_sibling_or_token_by_kind` tbh

src/api.rs
theo-lw: I suppose not. I'll change this

src/cursor.rs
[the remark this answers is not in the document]
theo-lw: Hmm, my observation is that using `first_child_or_token_by_kind` allows us to jump to the first child/token node that matches a kind without needing to allocate any other nodes. Contrast this to using `first_child_or_token` followed by `next_sibling_or_token_by_kind` (if the first child doesn't match the kind), which can potentially allocate the first child before jumping to the desired node. I think its this allocation avoidance that explains the improvement in the DHAT outputs.

src/api.rs
theo-lw: Changed

src/cursor.rs
theo-lw: Changed

milianw:
Hey there,

I've also run into this source of frequent allocations while profiling `slint` and wonder if I could help out in moving this patch forward?

Generally, could someone shed some light on _why_ we are allocating so much here in the first place? https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/syntax.md explains that the node data is held in an `Rc`, while in reality we have a `ptr: ptr::NonNull<NodeData>` in `SyntaxNode`. Still, the code like 
```
impl Clone for SyntaxNode {
    #[inline]
    fn clone(&self) -> Self {
        self.data().inc_rc();
        SyntaxNode { ptr: self.ptr }
    }
}
```
looks like it should be doing a simple refcount increment and not a deep copy. Yet heaptrack and gdb agree, this is not what happens in practice:
```
Breakpoint 2, 0x00007ffff64b3e58 in malloc () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff64b3e58 in malloc () from /usr/lib/libc.so.6
#1  0x000055555669fa0f in alloc::alloc::alloc (layout=...)
    at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/alloc.rs:98
#2  alloc::alloc::Global::alloc_impl (layout=..., zeroed=false, self=<optimized out>)
    at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/alloc.rs:181
#3  alloc::alloc::{impl#1}::allocate (layout=...) at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/alloc.rs:241
#4  alloc::alloc::exchange_malloc (size=64, align=8) at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/alloc.rs:330
#5  alloc::boxed::{impl#0}::new<rowan::cursor::NodeData> (x=<error reading variable: Cannot access memory at address 0x10>)
    at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/boxed.rs:257
#6  rowan::cursor::NodeData::new (parent=..., index=<optimized out>, offset=..., green=..., mutable=false) at src/cursor.rs:278
#7  0x000055555640fc09 in rowan::api::SyntaxNode<i_slint_compiler::parser::Language>::first_child<i_slint_compiler::parser::Language> (
    self=0x7fffffff6110) at /home/milian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.16/src/api.rs:149
#8  i_slint_compiler::parser::SyntaxNode::child_node (self=0x40, kind=i_slint_compiler::parser::SyntaxKind::DoubleArrow)
    at internal/compiler/parser.rs:812
```
The rust compiler seems to be missing a lot of inline debug frame information here, I cannot follow how it goes from
```
pub fn first_child(&self) -> Option<SyntaxNode<L>> {
    self.raw.first_child().map(Self::from)
}
```
straight to calling
```
impl NodeData {
    #[inline]
    fn new(
        parent: Option<SyntaxNode>,
        index: u32,
        offset: TextSize,
        green: Green,
        mutable: bool,
    ) -> ptr::NonNull<NodeData> {
```
which then does
```
                    it => {
                        let res = Box::into_raw(Box::new(res));
                        it.add_to_sll(res);
                        return ptr::NonNull::new_unchecked(res);
                    }
```
If this wouldn't allocate but just refcount, then the current way of iterating would become much faster without having to introduce new API?

milianw (changes requested):
`api.rs` is missing forwards for `first_child_by_kind`,  `first_child_or_token_by_kind`, `next_sibling_by_kind` and `next_sibling_or_token_by_kind`

Also note that the `by_kind` uses the raw `rowan::SyntaxKind`, instead of the `L::Kind`. Sadly we cannot easily remap here, since the `dyn Fn` defines the lifetime - which is generally quite a clutch to use this in practice. I.e. it's impossible to hand out iterators in our case:

```

    pub fn children_with_tokens_by_kind(
        &self,
        kind: SyntaxKind,
    ) -> impl Iterator<Item = NodeOrToken> {
        let source_file = self.source_file.clone();
        self.node
            .children_with_tokens()
            .by_kind(&|raw_kind| raw_kind == kind.into())
            .filter(move |n| n.kind() == kind)
            .map(move |token| match token {
                rowan::NodeOrToken::Node(node) => {
                    SyntaxNode { node, source_file: source_file.clone() }.into()
                }
                rowan::NodeOrToken::Token(token) => {
                    SyntaxToken { token, source_file: source_file.clone() }.into()
                }
            })
    }
```

this FTBFS with
```
error[E0373]: closure may outlive the current function, but it borrows `kind`, which is owned by the current function
   --> internal/compiler/parser.rs:840:23
    |
840 |             .by_kind(&|raw_kind| raw_kind == kind.into())
    |                       ^^^^^^^^^^             ---- `kind` is borrowed here
    |                       |
    |                       may outlive borrowed value `kind`
    |
note: closure is returned here
   --> internal/compiler/parser.rs:838:9
    |
838 | /         self.node
839 | |             .children_with_tokens()
840 | |             .by_kind(&|raw_kind| raw_kind == kind.into())
841 | |             .filter(move |n| n.kind() == kind)
...   |
848 | |                 }
849 | |             })
    | |______________^
help: to force the closure to take ownership of `kind` (and any other referenced variables), use the `move` keyword
    |
840 |             .by_kind(&move |raw_kind| raw_kind == kind.into())
    |                       ++++

error[E0716]: temporary value dropped while borrowed
   --> internal/compiler/parser.rs:840:23
    |
838 | /         self.node
839 | |             .children_with_tokens()
840 | |             .by_kind(&|raw_kind| raw_kind == kind.into())
    | |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
841 | |             .filter(move |n| n.kind() == kind)
...   |
848 | |                 }
849 | |             })
    | |______________- opaque type requires that borrow lasts for `'static`
850 |       }
    |       - temporary value is freed at the end of this statement
```

So basically, the `by_kind` code is unusable for our purpose - how do you envision to use this? Boxing the lambda callback is obviously a no-go as we want to elide allocations here. Is there something like C++ small-buffer optimized `std::function` in Rust which could be leveraged here instead?

milianw:
Regarding my last statement wrt `by_kind` being unusable - here is an approach that works for me: https://github.com/rust-analyzer/rowan/pull/171/commits/60a632ad984ab451e32058169193511154c675a9

PR #121, opened 2021-10-26
[DRAFT] - Try to optimize preorder iteration with pool allocation

theo-lw:
Preorder iteration allocates O(size of tree) nodes. However, there are some cases where only O(height of tree) nodes are alive at a given time. I'm trying to see if using a pool allocator can help avoid excess allocations in this case.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -193,6 +193,10 @@ impl<L: Language> SyntaxNode<L> {
         self.raw.descendants().map(SyntaxNode::from)
     }
 
+    pub fn descendants_pooled(&self, capacity: usize) -> impl Iterator<Item = SyntaxNode<L>> {
+        self.raw.descendants_pooled(capacity).map(SyntaxNode::from)
+    }
+
     pub fn descendants_with_tokens(&self) -> impl Iterator<Item = SyntaxElement<L>> {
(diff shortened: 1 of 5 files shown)


matklad:
Yeah, in general, it is a reasonable assumption that at any given point at time there are only few live nodes for any given tree. (although might be good to check that using rust-analyzer's analysis_stats). So it'd be cool to just add nodes to / pop them from a free-list. Ideally, tree traversal which doesn't retain nodes shouldn't do any allocation or atomic ops. 

In the original implementation, this was achieved by using a thread-local free-list of nodes. As it turned out several years later, at least my implementation of that apporoach was slower than just allocating the nodes.

theo-lw:
> In the original implementation, this was achieved by using a thread-local free-list of nodes. As it turned out several years later, at least my implementation of that apporoach was slower than just allocating the nodes.

Yeah right now my implementation isn't faster than just allocating the nodes. Maybe this is a fruitless endeavour.

theo-lw:
With my latest changes I do see a small speedup:

Without these changes:

Screen Shot 2021-10-31 at 6 45 26 PM

With these changes:

Screen Shot 2021-10-31 at 6 44 37 PM

PR #122, opened 2021-11-03
Add AstPtr and AstNode

azdavis:
This basically lifts from 

- https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/syntax/src/ptr.rs
- https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/syntax/src/ast.rs#L38-L64

I've found myself wanting `AstPtr` and `AstNode` in my own side projects and ended up basically copy-pasting them from rust-analyzer internals. Since these seem generic enough (not rust specific) maybe they could be here in rowan.

If this seems ok to merge I can follow up to migrate rust-analyzer to this.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -5,8 +5,8 @@ use crate::{
     SyntaxKind, SyntaxText, TextRange, TextSize, TokenAtOffset, WalkEvent,
 };
 
-pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
-    type Kind: fmt::Debug;
+pub trait Language: Sized + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
+    type Kind: Sized + Copy + fmt::Debug + Eq + Ord + std::hash::Hash;
 
     fn kind_from_raw(raw: SyntaxKind) -> Self::Kind;
     fn kind_to_raw(kind: Self::Kind) -> SyntaxKind;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
 description = "Library for generic lossless syntax trees"
-edition = "2018"
+edition = "2021"
 
 exclude = [".github/", "bors.toml", "rustfmt.toml"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.14.1"
+version = "0.15.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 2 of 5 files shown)


lnicola:
Seems reasonable, but please bump the version number too. Is the 2021 change required? It's arguable a breaking change so you should probably use `0.15` in that case.

r? @matklad

matklad:
Can we actually use these in rust-analyzer? As in, would it run in some coherence problems, or is it usable as is?

azdavis:
Let me first see if I can migrate rust analyzer to this locally before we merge this.

azdavis:
The 2021 version isn't required but I thought it'd be good to do anyway.

azdavis:
I tried migrating rust-analyzer to this version of rowan, see above.

azdavis:
The only downside I see to this is that it necessitates using `AstNode<Language = RustLanguage>` instead of the previous `AstNode` in rust-analyzer (and probably any similar crate that would use `AstNode`.

When migrating, I tried something like

```rs
pub trait AstNode: rowan::ast::AstNode<Language = RustLanguage> {}

impl<N: rowan::ast::AstNode<Language = RustLanguage>> AstNode for N {}
```

but then the methods on `rowan::ast::AstNode` are only in scope if you explicitly `use` it (`not` the rust-specific `AstNode`), which is a little confusing.

azdavis:
Thoughts on this? As I see it it’s a mild ergonomics hit (as opposed to not having to specify the language) but in return downstream Rowan users don’t have to invent their own versions of these (what i believe to be) fairly common things.

src/api.rs
```diff
@@ -5,8 +5,8 @@ use crate::{
     SyntaxKind, SyntaxText, TextRange, TextSize, TokenAtOffset, WalkEvent,
 };
 
-pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
-    type Kind: fmt::Debug;
+pub trait Language: Sized + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
+    type Kind: Sized + Copy + fmt::Debug + Eq + Ord + std::hash::Hash;
```
Veykril, line 9: Why do we need these extra bounds now?

Veykril:
> The only downside I see to this is that it necessitates using AstNode<Language = RustLanguage> 

Makes you wish we had trait aliases already

src/api.rs
Veykril: I see, the bounds make sense for this assoc type anyways I'd say so that's fine.

azdavis:
Is this ok to merge?

lnicola:
¯\\_(ツ)\_/¯

r? @matklad again

matklad:
Ok, so I think it seems useful for building smaller things on rowan, as you get more of the scaffolding out of the box. 

For rust-analyzer though, I think we shouldn't use `AstNode` from rowan, and instead should define our own AstNode. 

```rust
    where
        Def::Ast: AstNode<Language = RustLanguage>,
```

That `Language = RustLanguage` is accidental complexity which doesn't bring any benefit to rust-analyzer. As ast is the core vocabulary type, it should be as simple as possible, even if that means we need to duplicate some of rowan's code. 

Re-using `SyntaxNodePtr` should be fine I think, as that can be just an alias. 

bors r+

PR #129, opened 2022-02-15
Add ast support module, extracted from rust-analyzer

mullr:
I used this code from rust-analyzer in order to be able to build out my own AST structures. The only change I made was to make it generic over language.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -17,9 +17,10 @@ use std::{
     fmt,
     hash::{Hash, Hasher},
     iter::successors,
+    marker::PhantomData,
 };
 
-use crate::{Language, SyntaxNode, TextRange};
+use crate::{Language, SyntaxNode, SyntaxNodeChildren, TextRange};
 
 /// The main trait to go from untyped [`SyntaxNode`] to a typed AST. The
 /// conversion itself has zero runtime cost: AST and syntax nodes have exactly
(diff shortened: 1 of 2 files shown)


mullr:
Is there any desire to have this code in the main rowan repo? If not, I can make a separate utility crate. Just don't want to maintain an unnecessary fork.

src/ast.rs
```diff
@@ -162,3 +163,39 @@ impl<N: AstNode> From<AstPtr<N>> for SyntaxNodePtr<N::Language> {
...
+        parent.children().find_map(N::cast)
+    }
+
+    pub fn children<N: AstNode>(parent: &SyntaxNode<N::Language>) -> AstChildren<N> {
+        AstChildren::new(parent)
+    }
+
+    pub fn token<L: Language>(parent: &SyntaxNode<L>, kind: L::Kind) -> Option<SyntaxToken<L>> {
+        parent.children_with_tokens().filter_map(|it| it.into_token()).find(|it| it.kind() == kind)
+    }
```
Veykril, line 200: ~~Not entirely sure how useful the `token` will be given you'll have to specify the language type which seems kind of annoying.~~

Actually it should get inferred from the `parent` parameter.

Veykril (approved):
Sounds reasonable to me to have these exposed for a bit nicer typed handling of things.

Veykril:
Ah, can you bump the rowan version to `0.15.4`? Then merging will auto-publish the new minor version

mullr:
> Ah, can you bump the rowan version to 0.15.4? 

Done

PR #132, opened 2022-04-23
Rename `SyntaxToken::ancestors`

Veykril:
bors r+
(does bors work again?)


lnicola:
Oops, wrong repo?

Veykril:
bors.tech is still bugged on all of the rust-analyzer org it seems (@bors isn't set up here)

PR #135, opened 2022-07-09
missing re-export

### LINKED ISSUE #134: NodeCache not exposed
`GreenNodeBuilder` has a public `with_cache` constructor, but the `NodeCache` type is not public and I can't see any way to construct one outside of the library.

matklad:
closes #134


lnicola:
bors ping

I don't think it's working any more.

PR #140, opened 2022-09-15
#139 - Fix prev_sibling to actually return the previous sibling

### LINKED ISSUE #139: prev_sibling does not work as expected
A call to `syntax_node.prev_sibling()` returns actually the node two nodes before itself instead of the one before it.
I found out that `syntax_node.prev_sibling_or_token` works as expected and finds also the previous node.

To illustrate this issue, take the following tree of syntax nodes (only nodes, not tokens):
```
Root
    A
    B
    C
```
The calls to `prev_sibling` and `prev_sibling_or_token` report the following (per current node)
```
A
    prev_sibling=None
    prev_sibling_or_token=None
B
    prev_sibling=None
    prev_sibling_or_token=Some(Node(A))
C
    prev_sibling=Some(A)
    prev_sibling_or_token=Some(Node(B))
```

I found this by working with the rowan syntax trees in my rewrite of ( https://github.com/MalteJanz/ludtwig/tree/bachelor-thesis ) where I use rowan ( still in a very early state ). Parsing is done by the `GreenNodeBuilder` (abstracted away behind a sink / events). I have not yet found the time to produce a minimal reproducible example, so it would be nice if someone can confirm this behaviour in their rowan syntax trees or proof me wrong in my understanding of the function 🙈 .

I guess I already found the underlaying issue and will submit a PR shortly (offset by one error in [`src/cursor.rs` line 393](https://github.com/rust-analyzer/rowan/blob/188c607ff0c776d5c61fb58e0523ff110eb2c8c7/src/cursor.rs#L393)) 🙂.

MalteJanz:
closes #139

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.8"
+version = "0.15.9"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -390,7 +390,7 @@ impl NodeData {
     }
     fn prev_sibling(&self) -> Option<SyntaxNode> {
         let mut rev_siblings = self.green_siblings().enumerate().rev();
-        let index = rev_siblings.len() - (self.index() as usize);
+        let index = rev_siblings.len() - 1 - (self.index() as usize);
 
         rev_siblings.nth(index);
         rev_siblings.find_map(|(index, child)| {


matklad:
Can't believe no-one notice this before, excellent catch! Could you bump a patch version here to get a release https://github.com/rust-analyzer/rowan/blob/master/Cargo.toml#L3 ?

matklad:
@bors r+

Thanks! Also, sorry that the testing situation with roawn is suboptimal... 

I feel like we should do some small language in this repo, to have proper tests...

MalteJanz:
Done 👍 . A patch release would be nice. I was also quite suprised that my tree iteration doesn't work as expected when using the `prev_sibling` functions and that nobody encountered this before me😬.

By the way: you did an amazing job explaining how rowan works over here [syntax.md](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/syntax.md). I am still quite new to the whole parsing and syntax tree world and it was a good read😅.

matklad:
@Veykril do we have any bors here?

bjorn3:
Yes, but I believe it has been broken ever since rust-analyzer moved to the rust-lang org.

Veykril:
Yep, bors is broken all across rust-analyzer since the move ...

PR #147, opened 2022-12-26
making use of logos instead of m_lexer in #examples

dzmitry-lahoda:
because logos seems standard lexer in rust crates :) 

so this makes examples more approachable for newcomers and more production ready

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ countme = "3.0.0"
 serde = { version = "1.0.89", optional = true, default-features = false }
 
 [dev-dependencies]
-m_lexer = "0.0.4"
+logos = "0.12.1"
 
 [features]
 serde1 = [ "serde", "text-size/serde" ]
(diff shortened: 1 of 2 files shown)


matklad:
I’d rather avoid heavy deps here, my prefenrce would be to use either something super simple (m_lexer), or just to hand-write stuff.

dzmitry-lahoda:
logos is simple. i mean it is simplier then m_regex for me. easy to read. less code. no need to learn m_regex to understand rowan. and heavy dep is heavy in example, not in release.

matklad:
Logos is a proc macro, I want to avoid that.

dzmitry-lahoda:
i was thinking about some kind of flow, one may want to parse new language. it goes to logos, then to rowan, then to whatever integration on top. so examples is example of binding logos to rowan. unlikely people will use m_lexer. so this pr saves some time of this process.

Issue #144, opened 2022-10-20
Bug with SyntaxPtr::to_node + many nodes with same range


azdavis:
[`SyntaxNodePtr::to_node`][1] uses [`SyntaxNode::child_or_token_at_range`][2], whose docs state

> If there are several intersecting elements, any one can be returned.

This leads to issues (i.e. spurious panics) when there are indeed several elements that share the same range.

In my project, I have a [somewhat minimal reproducer][3]:

```
$ cargo t -p tests rowan_panic
    Finished test [unoptimized + debuginfo] target(s) in 0.13s
     Running unittests src/lib.rs (target/debug/deps/tests-bbc3b3005ec8d90f)

running 1 test
test misc::rowan_panic ... FAILED

failures:

---- misc::rowan_panic stdout ----
root: Root@0..17
  Whitespace@0..1 "\n"
  Dec@1..17
    DecWithTailInSeq@1..17
      DecWithTail@1..17
        DecInSeq@1..17
          FunctorDec@1..17
            FunctorKw@1..8 "functor"
            Whitespace@8..9 " "
            FunctorBind@9..17
              Name@9..16 "MkThing"
              Whitespace@16..17 "\n"
              Dec@17..17
                DecWithTail@17..17
    DecWithTail@17..17

=====
ptr: SyntaxNodePtr { kind: FunctorDec, range: 1..17 }
node: Root@0..17
node: Dec@1..17
node: DecWithTailInSeq@1..17
node: DecWithTail@1..17
node: DecInSeq@1..17
node: FunctorDec@1..17
resolved: FunctorDec@1..17
=====
ptr: SyntaxNodePtr { kind: Dec, range: 17..17 }
node: Root@0..17
node: Dec@1..17
node: DecWithTail@17..17
thread 'misc::rowan_panic' panicked at 'can't resolve local ptr to SyntaxNode: SyntaxNodePtr { kind: Dec, range: 17..17 }', crates/mlb-statics/src/lib.rs:358:24
```

The second ptr is to a Dec with range 17..17, which indeed exists, but it seems that this is not chosen for consideration and instead the DecWithTail with the same range is chosen instead. It does not have the same SyntaxKind, leading to the panic.

[1]: https://github.com/rust-analyzer/rowan/blob/295e5d03e822639b17144fcf51a4c3fb0ad774ed/src/ast.rs#L70-L88
[...]

PR #148, opened 2023-01-16
Add try_to_node which does not panic

azdavis:
https://github.com/rust-analyzer/rowan/issues/144 hasn't seen any movement and I'm not entirely sure how to fix it myself. So in the meantime it'd be nice to not panic and just fail more gracefully. It's also nice in general as a library to expose the possibility of handling errors, without panicking, to callers.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -70,7 +70,7 @@ impl<L: Language> SyntaxNodePtr<L> {
     /// Like [`Self::try_to_node`] but panics instead of returning `None` on
     /// failure.
     pub fn to_node(&self, root: &SyntaxNode<L>) -> SyntaxNode<L> {
-        self.try_to_node(root).unwrap()
+        self.try_to_node(root).unwrap_or_else(|| panic!("can't resolve {self:?} with {root:?}"))
     }
 
     /// "Dereferences" the pointer to get the [`SyntaxNode`] it points to.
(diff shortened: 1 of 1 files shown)


azdavis:
@Veykril could this have a look?

azdavis:
More discussion in the linked issue #144 but the original reason why i opened this pr has been fixed. If we don't think this is useful in general we can close, although it might be nice simply because it lets the caller choose whether they want to panic or not. As I understand, generally libraries should only panic when:

- Internal invariants are violated (a bug in the library code)
- "Obviously" incorrect arguments are passed

I'm not sure trying and failing to resolve a syntax node pointer fits into either case, so maybe we should allow for catching the error by returning an option without panicking via `try_to_node`.

azdavis:
Could we make a decision either way on this?

src/ast.rs
```diff
@@ -67,24 +67,30 @@ impl<L: Language> SyntaxNodePtr<L> {
...
     /// The complexity is linear in the depth of the tree and logarithmic in
     /// tree width. As most trees are shallow, thinking about this as
     /// `O(log(N))` in the size of the tree is not too wrong!
-    pub fn to_node(&self, root: &SyntaxNode<L>) -> SyntaxNode<L> {
-        assert!(root.parent().is_none());
-        successors(Some(root.clone()), |node| {
-            node.child_or_token_at_range(self.range).and_then(|it| it.into_node())
-        })
-        .find(|it| it.text_range() == self.range && it.kind() == self.kind)
-        .unwrap_or_else(|| panic!("can't resolve local ptr to SyntaxNode: {:?}", self))
```
Veykril, line 87: can we keep this unwrap message in the new `to_node` implementations?

PR #149, opened 2023-01-21
Bump `memoffset` and `hashbrown`

lnicola:
Helps with some duplication after https://github.com/rust-lang/rust-analyzer/pull/13996.


Veykril:
I think I already did this judging from the master copy of Cargo.toml

PR #151, opened 2023-02-18
Prevent use of SyntaxNodePtr and AstPtr on mutable trees

### LINKED ISSUE #150: AstPtr can de-sync from mutable Syntax Tree after mutation
If I keep `AstPtr`s to two nodes in a tree (appearing one after the other in source code), and make changes to the first one, the second `AstPtr` becomes invalid and will panic when `to_node` is called

In retrospect I should've seen this coming considering `AstPtr` only keeps a source code location reference to the node in the tree, and the tree source physically gets changed when mutations occur

Pseudocode to demonstrate the issue:
```rs
println!("Create Tree");
let mut tree = make_module("root");

println!("Add 2 children");
tree.append_child(make_module("child_1"));
tree.append_child(make_module("child_2"));

println!("Create child pointers");
let child_1_ptr = AstPtr::new(&tree.ast_children().nth(0).unwrap());
let child_2_ptr = AstPtr::new(&tree.ast_children().nth(1).unwrap());

println!("Resolve both pointers");
let mut child_1 = child_1_ptr.to_node(tree.syntax());
let child_2 = child_2_ptr.to_node(tree.syntax()); // All fine here, no mutations done

println!("Mutate tree under child 1");
child_1.append_child(make_module("child_1_1")).unwrap();

println!("Try to resolve pointer 2");
let child_2 = child_2_ptr.to_node(tree.syntax()); // Panic!
```

I'm fairly certain this is by design, but is there any way around this? If not, I could make a PR to add this as a note to the docs

Technohacker:
Both of these use source code locations to identify the node, which can get invalidated by syntax tree mutations. This commit adds assertions to prevent their use, and adds documentation to inform users of the issue

Fixes #150

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -247,6 +247,10 @@ impl<L: Language> SyntaxNode<L> {
         SyntaxNode::from(self.raw.clone_for_update())
     }
 
+    pub fn is_mutable(&self) -> bool {
+        self.raw.is_mutable()
+    }
+
     pub fn detach(&self) {
         self.raw.detach()
     }
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -55,6 +55,11 @@ pub trait AstNode {
 }
 
 /// A "pointer" to a [`SyntaxNode`], via location in the source code.
+///
+/// ## Note
+/// Since the location is source code dependent, this must not be used
+/// with mutable syntax trees. Any changes made in such trees causes
+/// the pointed node's source location to change, invalidating the pointer.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub struct SyntaxNodePtr<L: Language> {
     kind: L::Kind,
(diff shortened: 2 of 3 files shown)


Technohacker:
There's an added `is_mutable()` method for `cursor::SyntaxNode` and `api::SyntaxNode` which was needed for use in the API module. It may also be useful as a public API in case users wish to check if a tree is mutable

There's also 2 tests added to ensure this panic behaviour, but I've had to make a custom impl of `Language` to do so

jelmer:
@Technohacker this appears conflicted now

Technohacker:
Ah I can get that resolved in a few mins

Technohacker:
All done :+1: 
I've still assumed mutable trees should cause panics since I figured it's a programming mistake to use SyntaxNodePtrs on mutable trees. Should it be an Option return like the missing node case instead?

Also not sure what's going on with the CI, could you have a check?

jelmer:
Hmm, it appears to fail in ``master`` for the same reasons. https://github.com/rust-analyzer/rowan/actions/runs/9030330328/job/24814372618

Veykril:
Well that's a fun new lint we trigger

Veykril (approved):

PR #158, opened 2023-06-28
Bump version

azdavis:
Now that #148 is in I'd like to have a new version so I can grab it.

It does look like CI is failing recently though. Maybe the token is expired? Or it needs to be the new format of crates.io tokens? https://blog.rust-lang.org/2023/06/23/improved-api-tokens-for-crates-io.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.12"
+version = "0.15.13"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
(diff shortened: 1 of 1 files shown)


Veykril:
matklad recently revoked all of their rust-analyzer related tokens which is why the publish is failing here, we'd need to add the rust lang publisher thing as an owner for rowan as well I suppose

azdavis:
Can @matklad help with that?

azdavis:
any progress on this?

Veykril:
Looks like the owner was added, will bump someone from the infra team tomorrow to add the token to the secrets

azdavis:
how did that go?

PR #161, opened 2024-01-21
Use niching for Checkpoint

### LINKED ISSUE #160: Niching for Checkpoint
Since Checkpoint is a fairly arbitrary integer newtype, it would be nice to use the NonZero* variant of the internal integer type to allow niching. I think it would be fairly simple to do: just add one when wrapping the type and subtract one when unwrapping it.

mattfbacon:
Fixes #160.


Veykril:
Seems reasonable to me

PR #164, opened 2024-05-31
Remove `memoffset` dependency.

aDotInTheVoid:
`std::mem::offset` has been availible on stable since 1.77.0. `memoffset` uses this when compiled against a recent enough rustc.

Moving to the implementation in `std` lets me drop both `memoffset` and `autocfg` from my transitive dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.15"
+version = "0.15.16"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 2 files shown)


lnicola:
Since we're here, we should probably add a `rust-version` key. You can ignore the CI error, it's a false positive.

Veykril:
And bump the version number in the Cargo.toml so CI does a release

aDotInTheVoid:
@Veykril do you want a minor or patch release? this crate doesn't have an MSRV policy.

Veykril:
Patch should be fine, I don't see any breaking changes isnce the last bump

aDotInTheVoid:
Done

Veykril:
Thanks!

aDotInTheVoid:
Gah, the autopublish doesn't work due to false positives in `invalid_reference_casting`. They're fixed on nightly, but not beta (but are backport nominated). I don't need this on crates.io anytime soon, so I'm happy to wait for CI to become unborked, but it'll probably require a manual CI rerun of the specific commit (if you want `v0.15.16` to correspond to that commit)

PR #165, opened 2024-06-21
Allow passing an iterable to `splice_children`

michaelvanstraten:
Should reduce two allocation when splicing children.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.16"
+version = "0.16.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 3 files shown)


lnicola (approved):
Can you also bump the crate version number? I guess this isn't a breaking change.

michaelvanstraten:
@lnicola, sorry for the delay, I've bumped the version number.

milianw:
@Veykril can you also review and merge this? I assume bumping the crate version would also have been required for my PR, so getting this in here instead would solve this too.

Veykril:
This needs a minor version bump, the signature change here could break callers

michaelvanstraten:
> This needs a minor version bump, the signature change here could break callers

I don't see exactly how, since `I: IntoIterator<Item = SyntaxElement<L>>` is implemented for  `Vec<SyntaxElement>`.

michaelvanstraten:
@Veykril is updated.

milianw:
@Veykril ping? can this be merged now and a new release be tagged?

PR #171, opened 2024-10-27
Optimize children iteration by reusing NodeData if possible

milianw:
The first patch here is a rebased and slightly adapted version of https://github.com/rust-analyzer/rowan/pull/119 (the adaption is the addition of missing forwards in `api.rs`). I believe that part should be reviewed in the existing pull request which hopefully could be updated by @theo-lw (feel free to take the patch here if it's to your likings).

The second patch in here is new - it's an additional optimization which allows us to further reduce the number of memory allocations required for rowan tree visitation in `slint` by another 50%. I.e. combined, the two patches go from 850k allocations down to 450k down to 230k, a total reduction by ~73%!


src/cursor.rs
```diff
@@ -564,6 +597,19 @@ impl SyntaxNode {
...
+    #[inline]
+    fn can_take_ptr(&self) -> bool {
+        self.data().rc.get() == 1 && !self.data().mutable
+    }
+
+    #[inline]
+    fn take_ptr(self) -> ptr::NonNull<NodeData> {
+        assert!(self.can_take_ptr());
+        // increment the refcount before self gets dropped which will decrement the refcount again
+        self.data().inc_rc();
```
Veykril, line 609: Couldn't we just `mem::forget(self)` instead then?

src/cursor.rs
```diff
@@ -853,6 +988,19 @@ impl SyntaxToken {
...
```
Veykril, line 1000: likewise

src/cursor.rs
milianw: done, TIL

PR #174, opened 2024-11-07
Implement Eq and Hash for GreenTokenData

hurryabit:
`GreenToken` implements all traits required to use it as the key type of a hash map. It also implements `Borrow<GreenTokenData>`. Thus is would be nice if `GreenTokenData` also implemented `Eq` and `Hash` such that the result of `SyntaxToken::green` could be used for lookups in such a hash map without additional overhead in terms of syntax or runtime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.15"
+version = "0.15.16"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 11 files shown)


matklad:
Consider if you might want to intern GreenTokens&Nodes, and use pointer-equality and identity instead of content hashing

hurryabit:
Thanks for your quick response.

That's basically what I'm doing. I'm still prototyping and wanted to use the `GreenToken`s as the keys of my interning table instead of putting an interner into the lexer.

I just realised that the `GreenNodeBuilder` already does interning but it's unclear to me how to leverage that for a lookup table with the current API.

BenjaminBrienen:
What is the status of this? It received a review, but the necessary action isn't clear to me.

hurryabit:
I found a different way to solve my problem (I don't remember how though) and have given up on this PR. Shall I close it or would you be interested in this feature?

BenjaminBrienen:
I don't need it. Closing it is probably fine. The general advise is to implement standard traits such as `Hash` when possible, but there seems to be a reason that it isn't valid in this scenario.

Issue #175, opened 2024-11-21
Rowan upgrade to 0.16.0 breaks `prev_sibling` behaviour


MalteJanz:
Discovered by https://github.com/MalteJanz/ludtwig/pull/122

The failed test pipeline contains a bunch of issues related to the rowan upgrade that might need further investigation. But I think the easiest failing test is the following:

[(can be found here)](https://github.com/MalteJanz/ludtwig/blob/09e8c5180bab611783939bd63536e05206ad8852/crates/ludtwig-parser/src/lib.rs#L135)
```rust

    #[test]
    fn it_should_not_panic_on_prev_sibling_call() {
        let parse = parse("<div>a<hr/></div>");
        let root = SyntaxNode::new_root(parse.green_node);
        // println!("{}", debug_tree(&root));

        let prev = root.prev_sibling();
        // println!("prev sibling: {:?}", prev);
        assert!(prev.is_none());

        let child: HtmlTag = support::child(&root).unwrap();
        let prev = child.syntax().prev_sibling();
        // println!("{:?} prev sibling: {:?}", child, prev);
        assert!(prev.is_none());

        let child: HtmlTag = support::child(child.body().unwrap().syntax()).unwrap();
        let prev = child.syntax().prev_sibling();
        // println!("{:?} prev sibling: {:?}", child, prev);
        assert!(prev.is_some());
    }
```
Btw: I wrote this test while fixing my previous rowan contribution, see https://github.com/rust-analyzer/rowan/pull/143 . It basically came down to a off-by-one error which might regressed with the latest version, see initial issue: https://github.com/rust-analyzer/rowan/issues/139

Test output:
```
---- tests::it_should_not_panic_on_prev_sibling_call stdout ----
thread 'tests::it_should_not_panic_on_prev_sibling_call' panicked at crates/ludtwig-parser/src/lib.rs:152:9:
assertion failed: prev.is_some()
```

Which points to the last assertion and last line of the test above. Seems like the sibling somehow got lost 🤔 

### Test explaination
The parsing input:
```html
<div>a<hr/></div>
```
produces the following tree:
```
ROOT@0..17
  HTML_TAG@0..17
    HTML_STARTING_TAG@0..5
      TK_LESS_THAN@0..1 "<"
      TK_WORD@1..4 "div"
      HTML_ATTRIBUTE_LIST@4..4
      TK_GREATER_THAN@4..5 ">"
    BODY@5..11
      HTML_TEXT@5..6
[...]

PR #176, opened 2024-11-22
Fix prev_sibling indexing off-by-one

milianw:
The patch f06a2c903caa270e977a73ef5b9091f99e8fbe6c changed the code to use skip instead of nth, which lead to an off-by-one bug that was uncovered by unit tests in ludtwig, see [1].

[1]: https://github.com/MalteJanz/ludtwig/pull/122

Fixes: https://github.com/rust-analyzer/rowan/issues/175

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.16.0"
+version = "0.16.1"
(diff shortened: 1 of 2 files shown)


lnicola (approved):

lnicola:
Ouch, thanks!

PR #179, opened 2025-01-19
Update deps

azdavis:
rustc-hash and hashbrown updated, also some table formatting for complex deps

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,14 +13,19 @@ exclude = [".github/", "bors.toml", "rustfmt.toml"]
 members = ["xtask"]
 
 [dependencies]
-rustc-hash = "1.0.1"
-hashbrown = { version = "0.14.3", features = [
-    "inline-more",
-], default-features = false }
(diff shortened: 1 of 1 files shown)


lnicola:
We can bump `xshell` too, see https://github.com/matklad/xshell/pull/97.

azdavis:
there is no xshell to bump 🤔

lnicola:
Oh, we're using `xaction`.

matklad:
>Oh, we're using xaction.

FWIW, I know consider xaction to be a bad idea, and prefer to directly run the relevant commands, copy-pasting the small section helper:

https://github.com/matklad/arbtest/blob/master/xtask/src/main.rs

azdavis:
can we bump the bumpable deps for now and maybe follow up to remove xaction if needed/desired?

azdavis:
looks like this was done in 42cb260f9bf7c7475804dbee01bbe616dfe604d3

PR #183, opened 2025-02-14
Make GreenChild use a defined layout

kornelski:


Veykril:
Thanks!
And thanks to [BenjaminBrienen](https://github.com/BenjaminBrienen) as well of course :)

PR #186, opened 2025-02-22
Update all deps, fix clippy, rust 2024

BenjaminBrienen:
Drops unmaintained dependency and implements it internally

### DIFF
--- a/examples/math.rs
+++ b/examples/math.rs
@@ -32,6 +32,7 @@ enum SyntaxKind {
     OPERATION,
     ROOT,
 }
+
 use SyntaxKind::*;
 
 impl From<SyntaxKind> for rowan::SyntaxKind {
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -4,7 +4,7 @@ use std::{
     cmp::Ordering,
     hash::{Hash, Hasher},
     marker::PhantomData,
-    mem::{self, offset_of, ManuallyDrop},
+    mem::{self, ManuallyDrop, offset_of},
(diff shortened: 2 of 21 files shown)


BenjaminBrienen:
@IoannisNezis @lnicola Can I get a review? I'm not sure if there is a rust-analyzer/rowan discord or zulip where I can ask about development (I don't see any thing in the readme), so I hope you don't mind the ping here.

Veykril:
Zulip for rowan would be rust-analyzer zulip, we don't really tend much to rowan anymore these days (as we plan on migrating away from it someday)

examples/math.rs
```diff
@@ -20,18 +20,18 @@ use std::iter::Peekable;
...
-
-    NUMBER,
-    ERROR,
-    OPERATION,
-    ROOT,
+    Number = 5,
+    Error = 6,
+    Operation = 7,
+    Root = 8,
 }
```
Veykril, line 33: Can we do undo this, the lint is allowed on top of it for a reason, the casing is intentiuonal as these are meant to be used unprefixed mimicking a bag of constants

src/arc.rs
```diff
@@ -312,10 +313,9 @@ impl<H, T> ThinArc<H, T> {
         };
 
         // Expose the transient Arc to the callback, which may clone it if it wants.
-        let result = f(&transient);
 
         // Forward the result.
```
Veykril, line 317:
```suggestion
```

Veykril (approved):

PR #187, opened 2025-02-23
Add more CI checks

BenjaminBrienen:
I mostly want to highlight what is currently failing in terms of lints and especially miri. This could incorporated into xtask.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,11 @@ on:
   pull_request:
   push:
     branches: ["master", "staging", "trying"]
+  merge_group:
+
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
I don't know how to fix the miri failure myself, but it is currently unsound and does affect downstream crates.

lnicola:
We'll probably rewrite `rowan` in the short/medium, we can resurrect this again after that. And one stylistic remark: in rust-analyzer, we use `rustup` instead of `dtolnay/rust-toolchain` and https://github.com/swatinem/rust-cache instead of `actions/cache`.

PR #209, opened 2026-03-24
feat: make no-std compatible

bitwalker:
This PR makes `rowan` no-std + liballoc compatible.

Currently, it also inlines the `countme` and `text_size` crates into respective modules under `src`, as they are also `std`-bound when they don't need to be. Those can be removed once those modules are made no-std compatible.

I'm opening this since it is easy to do so, but given the current maintenance status, I'm not expecting that it is likely to be merged - that said, just ping me if you want any changes, and I'm happy to make them, or to rebase this on a newer branch if needed.

I plan to publish a fork of this crate containing these changes to unblock my own use case, but would like to stick with mainline if possible, so if this does get merged (or the rewrite is no-std + liballoc compatible), I'll retire that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.15"
+version = "0.15.16"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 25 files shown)


ChayimFriedman2:
First, this will not be done before we complete the rewrite of `rowan`. After that...

> I'm opening this since it is easy to do so

I **heavily** object to that. Making a crate `no_std` just "since it is easy to do" is absolutely not a satisfying reason. `rowan` is a syntax tree library, something to be used by LSPs and the like, not things that usually run on bare metal or something.

So, if you have no use-case for this, I reject this. If you do (and please tell what it is), I might accept this depending on the changes it requires (after the rewrite, as I said).

Issue #192, opened 2025-04-06
MIRI UB in rowan `Arc::drop`


dnbln:
```
error: Undefined Behavior: write access through <3644568> at alloc1395441[0x0] is forbidden
   --> /Users/dinu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rowan-0.16.1/src/arc.rs:162:12
    |
162 |         if self.inner().count.fetch_sub(1, Release) != 1 {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ write access through <3644568> at alloc1395441[0x0] is forbidden
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: the accessed tag <3644568> has state Frozen which forbids this child write access
help: the accessed tag <3644568> was created here, in the initial state Frozen
```

bjorn3:
Does it work with `MIRIFLAGS="-Zmiri-tree-borrows"`?

dnbln:
That was from when I ran it with `-Zmiri-tree-borrows`. It also fails with Stacked Borrows, although in a different place

Issue #163, opened 2024-05-13
Undefined Behavior


Veykril:
```

  error: casting references to a bigger memory layout than the backing allocation is undefined behavior, even if the reference is unused
     --> src/green/node.rs:197:35
      |
  195 |         let repr: &Repr = &self.ptr;
      |                            -------- backing allocation comes from here
  196 |         unsafe {
  197 |             let repr: &ReprThin = &*(repr as *const Repr as *const ReprThin);
      |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: casting from `ThinArc<GreenNodeHead, GreenChild>` (8 bytes) to `HeaderSlice<GreenNodeHead, [GreenChild; 0]>` (16 bytes)
      = note: `#[deny(invalid_reference_casting)]` on by default
  
  error: casting references to a bigger memory layout than the backing allocation is undefined behavior, even if the reference is unused
     --> src/green/token.rs:139:35
      |
  138 |             let repr: &Repr = &self.ptr;
      |                                -------- backing allocation comes from here
  139 |             let repr: &ReprThin = &*(repr as *const Repr as *const ReprThin);
      |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: casting from `ThinArc<GreenTokenHead, u8>` (8 bytes) to `HeaderSlice<GreenTokenHead, [u8; 0]>` (16 bytes)
  
  error: could not compile `rowan` (lib) due to 2 previous errors
  ```

saethlin:
FYI this is a false positive in the lint. I fixed this in https://github.com/rust-lang/rust/pull/124978 but did not realize at the time that it should have been backported.

Veykril:
So while that fixed the UB reported here, there is still another UB popping up now

```text

error: Undefined Behavior: trying to retag from <212251> for SharedReadWrite permission at alloc69839[0x0], but that tag does not exist in the borrow stack for this location
   --> src\arc.rs:71:18
    |
71  |         unsafe { &*self.ptr() }
    |                  ^^^^^^^^^^^^
    |                  |
    |                  trying to retag from <212251> for SharedReadWrite permission at alloc69839[0x0], but that tag does not exist in the borrow stack for this location
    |                  this error occurs as part of retag at alloc69839[0x0..0x18]
    |
[...]

PR #210, opened 2026-04-05
fix: reduce Miri UB in GreenNode/GreenToken deref and HeaderSlice

avrabe:
## Problem

Miri flags undefined behavior under both stacked borrows and tree borrows:

- `GreenNode::deref` and `GreenToken::deref` create `&Repr` (fat `HeaderSlice<H, [T]>`) via `ThinArc::deref`, then cast to `&ReprThin` (thin `HeaderSlice<H, [T; 0]>`) — creating a reference to a smaller type from a larger one
- `HeaderSlice<H, [T; 0]>::deref` takes `&self` whose provenance only covers the `[T; 0]` part, then creates a fat `&HeaderSlice<H, [T]>` that exceeds that provenance

Related: #192, #163, #108

## Fix

**Minimal changes to avoid the specific UB paths:**

1. `GreenNode::deref` / `GreenToken::deref`: Access the `ThinArc`'s inner pointer directly via `ptr::addr_of!` instead of going through the fat→thin reference chain
2. `HeaderSlice<H, [T; 0]>`: Extract `as_fat_ref` helper that takes a raw `*const Self` to avoid inheriting `&self`'s narrow provenance
3. `ThinArc::ptr` made `pub(crate)` for direct access from green node/token

## What this fixes

The `node_cache` and basic green tree construction paths now pass Miri cleanly. This covers the common usage pattern (parse → build green tree → query).

## What this doesn't fix

The cursor iteration path (`SyntaxText`, `PreorderWithTokens`) and mutable tree operations (`clone_for_update`) still have provenance issues from deeper in the cursor infrastructure. These need more extensive changes.

## Testing

- `cargo test` — all tests pass
- `cargo +nightly miri test -- --skip ast --skip syntax_text --skip tidy` — passes clean (3/6 lib tests)

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -252,13 +252,26 @@ impl<H, T> HeaderSlice<H, [T]> {
     }
 }
 
+impl<H, T> HeaderSlice<H, [T; 0]> {
+    /// Get a fat reference to the full HeaderSlice from raw pointer.
+    ///
+    /// # Safety
+    /// `ptr` must point to a valid `HeaderSlice<H, [T; 0]>` that was allocated
+    /// with space for `length` elements of `T` after the header.
+    pub(crate) unsafe fn as_fat_ref<'a>(ptr: *const Self) -> &'a HeaderSlice<H, [T]> {
+        let len = unsafe { (*ptr).length };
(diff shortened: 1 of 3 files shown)


avrabe:
Superseded by a more complete fix on the fix/miri-soundness-v2 branch — will open a new PR.

PR #211, opened 2026-04-06
fix: address Miri UB in Arc, ThinArc, green types, and cursor

avrabe:
Hi — we use rowan in two safety-critical projects (rivet, spar) and ran into the Miri UB documented in #192, #163, #108. We took a shot at fixing it and would appreciate your feedback on the approach.

## What this changes

The core issue is that Miri's borrow models (stacked and tree borrows) flag provenance violations when:
- References to thin types (`HeaderSlice<H, [T; 0]>`) are used to access data beyond the thin layout
- `&NodeData` containing Cell fields freezes the allocation tag, preventing later deallocation
- Refcount access via `Cell::get()` or `&ArcInner` creates references that narrow provenance

The fix touches 4 files (+152/-63 lines):
- **arc.rs**: refcount access via raw pointers instead of `&ArcInner` references; ThinArc clone/drop without transient Arc
- **green/node.rs**: `GreenNodeData` wraps fat `Repr` (unsized) so provenance covers the full slice; `into_raw` extracts pointer from ThinArc directly
- **green/token.rs**: `text()` via raw pointer arithmetic; `into_raw`/`deref` via ThinArc pointer
- **cursor.rs**: SyntaxNode/Token ptr in `UnsafeCell`; `NodeData::self_alloc` field preserves original allocation provenance for deallocation; refcount decrement via raw `*mut u32`

## What we DON'T know

- **Performance impact**: We haven't benchmarked. The `self_alloc` field adds 8 bytes per NodeData. The UnsafeCell wrapper and raw pointer access patterns might affect optimization. We'd really appreciate if you could run your benchmarks.
- **Mutable tree path**: The `clone_for_update` tests (`ensure_mut_*`) still fail under Miri. Our fix focuses on the immutable tree path. The mutable path has additional provenance issues in the singly-linked list infrastructure.
- **Whether this is the right approach**: We chose minimal, targeted changes over a larger refactor. There might be cleaner designs (e.g., making the cursor fully raw-pointer-based). We're open to feedback.

## Testing

- `cargo test --lib` — 6/6 pass
- `cargo +nightly miri test --lib -- --skip ensure_mut` — 4/4 pass under `-Zmiri-tree-borrows`
- Under stacked borrows, the immutable path still has issues due to the stricter model
- Tested in downstream: rivet (26 yaml_cst tests pass Miri), spar

Happy to iterate on this based on your feedback. Thank you for maintaining rowan — it's a great library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.15"
+version = "0.15.16"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.16.1"
+version = "0.16.2"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.15.16"
+version = "0.16.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rowan"
-version = "0.16.0"
+version = "0.16.1"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 repository = "https://github.com/rust-analyzer/rowan"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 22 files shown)


phall1:
tested this against a NodeCache rehash repro from cyrs (cypher-frontend). arc/cursor changes here help, but `GreenTokenData::text` still hits SB UB at `src/green/token.rs:104` because only `GreenNodeData` got promoted to a DST in this PR. `GreenToken::deref` still hands out a narrow `&GreenTokenData` whose retag covers only `sizeof::<ReprThin>()`, so `addr_of!(self.data.slice) + from_raw_parts` trips:

```
error: Undefined Behavior: trying to retag from <N> for SharedReadOnly permission at allocM[0x18]
 --> src/green/token.rs:104:25
     let bytes = std::slice::from_raw_parts(slice_start, len);
```

cyrs CI hits it twice deterministically:
- https://github.com/phall1/cyrs/actions/runs/25638947905 (job 75256239232)
- https://github.com/phall1/cyrs/actions/runs/25639698954 (job 75257867300)

opened #212 with the same DST promotion for tokens. with both stacked the cyrs repro passes. self-contained miri test: https://github.com/phall1/rowan/blob/fix/sb-node-cache-rehash/tests/miri_node_cache_rehash.rs

one more pre-existing SB hit on this branch for the record: `ast::tests::ensure_mut_panic_on_create` -> `GreenNodeData::to_owned` -> `from_raw` -> `ThinArc::clone` -> count `fetch_add`. narrow `&GreenNodeData` doesn't cover the count field. separate fix from #212.

ChayimFriedman2:
We're about to rewrite rowan anyway and the new code should pass Miri (and is also a change of pretty much all code).

Thanks for your interest in contributing!

PR #212, opened 2026-05-10
fix(token): make GreenTokenData a DST to widen retag past slice tail

phall1:
stacks on #211. tokens still need the same DST treatment you gave nodes there.

with #211 alone, `GreenToken::deref` returns a narrow `&GreenTokenData` whose retag covers only `sizeof::<ReprThin>()`. `text()` then widens past that with `addr_of!(self.data.slice) + from_raw_parts`, which SB rejects:

```
error: Undefined Behavior: trying to retag from <N> for SharedReadOnly permission at allocM[0x18]
 --> src/green/token.rs:104:25
     let bytes = std::slice::from_raw_parts(slice_start, len);
```

real-world path: `hashbrown::RawTable` rehash in `NodeCache::token` -> `token_hash` -> `text()`. 

caught in a cypher frontend project im doing. 

- https://github.com/phall1/cyrs/actions/runs/25638947905 (miri/cyrs-fmt)
- https://github.com/phall1/cyrs/actions/runs/25639698954 (miri/cyrs-plan)

what changes: same shape as your node fix in #211. `GreenTokenData::data` becomes `HeaderSlice<H, [u8]>`, `deref` goes through `&self.ptr`, `into_raw`/`from_raw` use `thin_to_thick`, `text()` collapses to `from_utf8_unchecked(self.data.slice())`. one file, ~28 lines.

test: `tests/miri_node_cache_rehash.rs` builds 200 distinct tokens to force a rehash. fails on #211 alone, passes here. `cargo test` and `cargo +nightly miri test --test miri_node_cache_rehash` both green.

one more SB hit out of scope here for the record: `ast::tests::ensure_mut_panic_on_create` -> `ThinArc::clone` -> `count_ptr.fetch_add` derefs through narrow `&GreenNodeData` and loses the count's provenance. separate fix.

refs #163, #192, #108

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -108,7 +108,8 @@ impl<T: ?Sized> Clone for Arc<T> {
         // another must already provide any required synchronization.
         //
         // [1]: (www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html)
-        let old_size = self.inner().count.fetch_add(1, Relaxed);
+        let count_ptr = unsafe { ptr::addr_of!((*self.ptr()).count) };
+        let old_size = unsafe { (*count_ptr).fetch_add(1, Relaxed) };
 
         // However we need to guard against massive refcounts in case someone
         // is `mem::forget`ing Arcs. If we don't do this the count can overflow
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
Sorry for you investing time in this, but we'll rewrite rowan anyway as soon as we finish with the rust-analyzer assist migration, a rewrite that will make it much more performant and also likely to fix this issue (and even if not, the code this PR changes will be stale). So I'm inclined to close this PR, as well as https://github.com/rust-analyzer/rowan/pull/211.

Also note that rowan should pass Tree Borrows, and that this specific rule from Stacked Borrows is unlikely to make it into the final aliasing model for Rust.

phall1:
@ChayimFriedman2  --

no worries! I'm happy to close this. I should be following rust analyzer more closely -- 

Where's the best spot to subscribe if you don't mind pointing me in the right direction?

ChayimFriedman2:
[#t-compiler/rust-analyzer > Future Rowan](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Future.20Rowan/with/592524643), https://github.com/rust-lang/rust-analyzer/issues/15710, https://github.com/rust-lang/rust-analyzer/issues/18285, we also have a GSoC project for that.

PR #217, opened 2026-08-02
Remove mutable api

bit-aloo:
This time to master


bit-aloo:
I think we lost some portion history here, like all of v0.16 commit history.

ChayimFriedman2:
Maybe, but I don't care much: most of it was changes to mutable APIs.

PR #219, opened 2026-08-14
Make trivia first class member of token

bit-aloo:
Still making changes to RA

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -261,6 +261,10 @@ impl<L: Language> SyntaxToken<L> {
         self.raw.text_range()
     }
 
+    pub fn text_range_including_trivia(&self) -> TextRange {
+        self.raw.text_range_including_trivia()
+    }
+
     pub fn index(&self) -> usize {
         self.raw.index()
     }
--- a/src/api.rs
+++ b/src/api.rs
@@ -62,17 +62,28 @@ impl<L: Language> fmt::Display for SyntaxNode<L> {
 impl<L: Language> fmt::Debug for SyntaxToken<L> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:?}@{:?}", self.kind(), self.text_range())?;
-        if self.text().len() < 25 {
-            return write!(f, " {:?}", self.text());
-        }
         let text = self.text();
-        for idx in 21..25 {
-            if text.is_char_boundary(idx) {
-                let text = format!("{} ...", &text[..idx]);
-                return write!(f, " {:?}", text);
-            }
+        if text.len() < 25 {
+            write!(f, " {text:?}")?;
+        } else {
+            let idx = (21..25).find(|&idx| text.is_char_boundary(idx)).unwrap();
+            write!(f, " {:?}", format!("{} ...", &text[..idx]))?;
         }
-        unreachable!()
+        write!(f, " [")?;
+        for (idx, piece) in self.leading_trivia().enumerate() {
+            if idx > 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "{:?}({:?})", piece.kind(), piece.text())?;
+        }
+        write!(f, "] [")?;
+        for (idx, piece) in self.trailing_trivia().enumerate() {
+            if idx > 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "{:?}({:?})", piece.kind(), piece.text())?;
+        }
+        write!(f, "]")
     }
 }
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -273,6 +273,22 @@ impl<L: Language> SyntaxToken<L> {
         self.raw.text()
     }
 
+    pub fn text_including_trivia(&self) -> String {
+        self.raw.text_including_trivia()
+    }
+
+    pub fn leading_trivia(
+        &self,
+    ) -> impl DoubleEndedIterator<Item = SyntaxToken<L>> + ExactSizeIterator {
+        self.raw.leading_trivia().map(SyntaxToken::from)
+    }
+
+    pub fn trailing_trivia(
+        &self,
+    ) -> impl DoubleEndedIterator<Item = SyntaxToken<L>> + ExactSizeIterator {
+        self.raw.trailing_trivia().map(SyntaxToken::from)
+    }
+
     pub fn green(&self) -> &GreenTokenData {
         self.raw.green()
     }
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -139,20 +139,12 @@ unsafe fn free(mut data: ptr::NonNull<NodeData>) {
 impl NodeData {
     #[inline]
     fn new(
-        parent: Option<SyntaxNode>,
+        parent: Option<ptr::NonNull<NodeData>>,
         index: u32,
         offset: TextSize,
         green: Green,
     ) -> ptr::NonNull<NodeData> {
-        let parent = ManuallyDrop::new(parent);
-        let res = NodeData {
-            _c: Count::new(),
-            rc: Cell::new(1),
-            parent: parent.as_ref().map(|it| it.ptr),
-            index,
-            green,
-            offset,
-        };
+        let res = NodeData { _c: Count::new(), rc: Cell::new(1), parent, index, green, offset };
         unsafe { ptr::NonNull::new_unchecked(Box::into_raw(Box::new(res))) }
     }
 
@@ -296,8 +288,9 @@ impl SyntaxNode {
         index: u32,
         offset: TextSize,
     ) -> SyntaxNode {
+        let parent = ManuallyDrop::new(parent);
         let green = Green::Node { ptr: green.into() };
-        SyntaxNode { ptr: NodeData::new(Some(parent), index, offset, green) }
+        SyntaxNode { ptr: NodeData::new(Some(parent.ptr), index, offset, green) }
     }
 
     pub fn clone_subtree(&self) -> SyntaxNode {
@@ -560,8 +553,9 @@ impl SyntaxToken {
         index: u32,
         offset: TextSize,
     ) -> SyntaxToken {
+        let parent = ManuallyDrop::new(parent);
         let green = Green::Token { ptr: green.into() };
-        SyntaxToken { ptr: NodeData::new(Some(parent), index, offset, green) }
+        SyntaxToken { ptr: NodeData::new(Some(parent.ptr), index, offset, green) }
     }
 
     #[inline]
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -175,8 +175,10 @@ impl NodeData {
 
     #[inline]
     fn parent_node(&self) -> Option<SyntaxNode> {
-        let parent = self.parent()?;
-        debug_assert!(matches!(parent.green, Green::Node { .. }));
+        let mut parent = self.parent()?;
+        while !matches!(parent.green, Green::Node { .. }) {
+            parent = parent.parent()?;
+        }
         parent.inc_rc();
         Some(SyntaxNode { ptr: ptr::NonNull::from(parent) })
     }
@@ -197,10 +199,7 @@ impl NodeData {
     fn green_siblings(&self) -> slice::Iter<'_, GreenChild> {
         match &self.parent().map(|it| &it.green) {
             Some(Green::Node { ptr }) => unsafe { &*ptr.as_ptr() }.children().raw,
-            Some(Green::Token { .. }) => {
-                debug_assert!(false);
-                [].iter()
-            }
+            Some(Green::Token { .. }) => [].iter(),
             None => [].iter(),
         }
     }
@@ -558,6 +557,17 @@ impl SyntaxToken {
         SyntaxToken { ptr: NodeData::new(Some(parent.ptr), index, offset, green) }
     }
 
+    fn new_trivia(
+        green: &GreenTokenData,
+        parent: SyntaxToken,
+        index: u32,
+        offset: TextSize,
+    ) -> SyntaxToken {
+        let parent = ManuallyDrop::new(parent);
+        let green = Green::Token { ptr: green.into() };
+        SyntaxToken { ptr: NodeData::new(Some(parent.ptr), index, offset, green) }
+    }
+
     #[inline]
     fn data(&self) -> &NodeData {
         unsafe { self.ptr.as_ref() }
@@ -605,6 +615,10 @@ impl SyntaxToken {
                 ""
             }
         }
+    }
+
+    pub fn text_including_trivia(&self) -> String {
+        self.with_trivia().map(|token| token.text().to_owned()).collect()
     }
 
     #[inline]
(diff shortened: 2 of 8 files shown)


src/trivia.rs
```diff
@@ -0,0 +1,541 @@
+use std::{fmt, iter::FusedIterator, marker::PhantomData};
+
+use text_size::{TextRange, TextSize};
+
+use crate::{cursor, Language, SyntaxToken};
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
+pub enum TriviaPieceKind {
+    Newline,
```
ChayimFriedman2: It is not the job of rowan to define this. Trivias should just have a SyntaxKind. From rowan's point of view, they're tokens attached to tokens. The rest will be handled by rust-analyzer. This will also require fewer changes.

src/api.rs
```diff
@@ -269,6 +307,20 @@ impl<L: Language> SyntaxToken<L> {
         self.raw.text()
     }
 
+    pub fn text_trimmed(&self) -> &str {
```
ChayimFriedman2: I believe `text()` and `text_range()` should exclude trivia, and we should have a `text_including_trivia()` method (and ditto for `text_range()`). While I don't know for sure, code in r-a today does not handle trivia so I expect this to be easier.

src/green/token.rs
```diff
@@ -106,18 +110,66 @@ impl GreenTokenData {
...
+        (leading_len, trailing_len, total_len.into())
+    }
+
+    #[inline]
+    pub fn text_trimmed(&self) -> &str {
+        let (leading_len, trailing_len, total_len) = self.leading_trailing_total_len();
+
+        let start: usize = leading_len.into();
+        let end: usize = (total_len - trailing_len).into();
+        let text = unsafe { std::str::from_utf8_unchecked(self.data.slice()) };
```
ChayimFriedman2:
```suggestion
        let text = self.text();
```
Save the unsafe.

src/green/trivia.rs
```diff
@@ -0,0 +1,134 @@
+use std::{
```
ChayimFriedman2: I'm not pleased with the amount of public functions here. We're going to remove green trees so we need to minimize their usage. Things that do not *need* to be public should remain private.

src/api.rs
ChayimFriedman2: I also think it'll be easier if trivias will store their text (like regular tokens).

src/api.rs
```diff
@@ -246,6 +257,29 @@ impl<L: Language> SyntaxNode<L> {
 }
 
 impl<L: Language> SyntaxToken<L> {
+    pub fn new_detached<Leading, Trailing>(
```
ChayimFriedman2: Why do we need this? A token cannot be a root, this feels wrong.

src/cursor.rs
```diff
@@ -1000,4 +1045,131 @@ impl Iterator for PreorderWithTokens {
         next
     }
 }
-// endregion
+
+#[derive(PartialEq, Eq, Clone, Hash)]
+pub struct SyntaxTrivia {
```
ChayimFriedman2: I don't think we need this type. We can just have `fn leading_trivia(&self) -> impl Iterator<Item = SyntaxToken>` om `SyntaxToken`, and ditto for `trailing_trivia()`. Just store a list (even `Box<[GreenToken]>`) of leading trivia in a token, and ditto for trailing trivia.

ChayimFriedman2: In particular, using regular tokens for trivias simplifies both the implementation and the changes in r-a.

ChayimFriedman2 (changes requested):
I know you opened this as draft by I've still reviewed it so you don't waste time changing r-a to match an API we'll throw away.

src/trivia.rs
bit-aloo: Removed.

src/api.rs
bit-aloo: Yup, having both text and text_range with including trivia variants.

src/green/token.rs
bit-aloo: Thanks

src/green/trivia.rs
bit-aloo: Ah!! sorry, lazy me. Should be good now.

src/api.rs
bit-aloo: Removed. I was copy pasting from biome back then.

src/cursor.rs
bit-aloo: Its much better. Again, took inspiration from biome initially. Removed.

src/green/node.rs
```diff
@@ -124,9 +124,36 @@ impl GreenNodeData {
     /// Returns the length of the text covered by this node.
     #[inline]
     pub fn text_len(&self) -> TextSize {
```
ChayimFriedman2, line 126: When I said to not include trivia in the text, I meant for tokens. For nodes it seem reasonable to do include them - they're children (if indirect), and if we compare to the current state again, they'll likely be normal children of this node. Or do you have a reason to do differently?

src/api.rs
```diff
@@ -324,6 +375,13 @@ impl<L: Language> SyntaxElement<L> {
         }
     }
 
+    pub fn text_range_including_trivia(&self) -> TextRange {
```
ChayimFriedman2: And then `SyntaxElement` (and `GreenElement` won't have `text_range_including_trivia()` at all, because `SyntaxNode` does not have them).

src/cursor.rs
```diff
@@ -184,7 +184,9 @@ impl NodeData {
     #[inline]
     fn parent_node(&self) -> Option<SyntaxNode> {
         let parent = self.parent()?;
-        debug_assert!(matches!(parent.green, Green::Node { .. }));
+        if !matches!(parent.green, Green::Node { .. }) {
```
ChayimFriedman2: This seems too risky. Code will depend on the parent node of trivias to be the parent node of the parent token (also, I already said that I think it really is). So I think we should loop here. That said siblings of the parent token probably shouldn't be considered as siblings of us, that's what I think at least (but what is easier to implement in r-a?).

src/cursor.rs
```diff
@@ -334,7 +336,20 @@ impl SyntaxNode {
 
     #[inline]
     pub fn text_range(&self) -> TextRange {
-        self.data().text_range()
+        let full = self.data().text_range_including_trivia();
```
ChayimFriedman2: And here too I think it should return the full length. This is also a very common constant-time operation that you turned into linear.

src/cursor.rs
```diff
@@ -347,6 +362,22 @@ impl SyntaxNode {
         SyntaxText::new(self.clone())
     }
 
+    pub fn text_including_trivia(&self) -> SyntaxText {
+        SyntaxText::with_range(self.clone(), self.text_range_including_trivia())
+    }
+
+    pub fn leading_trivia(
```
ChayimFriedman2: And I don't think `SyntaxNode::leading_trivia()` and `SyntaxNode::trailing_trivia()` should exist (unless you have a specific reason to want them), we're attaching trivia to tokens and not nodes.

src/cursor.rs
```diff
@@ -564,6 +595,25 @@ impl SyntaxToken {
...
+        parent: SyntaxToken,
+        index: u32,
+        offset: TextSize,
+    ) -> SyntaxToken {
+        let parent = ManuallyDrop::new(parent);
+        let green = Green::Token { ptr: green.into() };
+        let data = NodeData {
+            _c: Count::new(),
+            rc: Cell::new(1),
+            parent: Some(parent.ptr),
```
ChayimFriedman2: So I think this is actually a better way to make the node the parent, just put it there to begin with (and use `new()`).

src/green/node.rs
```diff
@@ -124,9 +124,36 @@ impl GreenNodeData {
...
+        self.text_len_including_trivia() - self.leading_trivia_len() - self.trailing_trivia_len()
+    }
+
+    #[inline]
+    pub fn text_len_including_trivia(&self) -> TextSize {
         self.header().text_len
```
ChayimFriedman2, line 127: I think it should be the opposite: `text_len()` should be O(1) (the field), because it is more common.

src/cursor.rs
```diff
@@ -585,7 +635,14 @@ impl SyntaxToken {
 
     #[inline]
     pub fn text_range(&self) -> TextRange {
-        self.data().text_range()
+        let green = self.green();
+        let start = self.data().text_range_including_trivia().start() + green.leading_trivia_len();
```
ChayimFriedman2: Why not call `GreenToken::text_range()` directly?

src/cursor.rs
```diff
@@ -665,6 +730,18 @@ impl SyntaxToken {
                 .and_then(|element| element.last_token()),
         }
     }
+
+    pub fn leading_trivia(
+        &self,
+    ) -> impl DoubleEndedIterator<Item = SyntaxToken> + ExactSizeIterator {
+        SyntaxTriviaTokens::new(Some(self.clone()), true)
```
ChayimFriedman2: Why not some iterator combinator? Why create a new iterator type?

ChayimFriedman2 (changes requested):
Hope it won't be too hard for you to change r-a 🫢️

src/green/node.rs
bit-aloo: No, I needed them initially for the mechanical one-to-one translation from the current setup to having trivia embedded. I grepped through the current set of changes, and we’re not really using them anywhere now, so they can be removed.

The rule of thumb we’re following now is:
Tokens have two method variants: one with trivia and one without. Everything else nodes, elements, contains the full token snapshot, including trivia. If we need to make any changes around this, we are doing them directly in r-a.

src/cursor.rs
```diff
@@ -688,7 +776,7 @@ impl SyntaxElement {
     pub fn text_range(&self) -> TextRange {
         match self {
             NodeOrToken::Node(it) => it.text_range(),
-            NodeOrToken::Token(it) => it.text_range(),
+            NodeOrToken::Token(it) => it.text_range_including_trivia(),
```
ChayimFriedman2, line 742:
```suggestion
            NodeOrToken::Token(it) => it.text_range(),
```
I'm a bit on the edge about this but think this is more correct.

src/green/element.rs
```diff
@@ -74,7 +74,7 @@ impl GreenElementRef<'_> {
     pub fn text_len(self) -> TextSize {
         match self {
             NodeOrToken::Node(it) => it.text_len(),
-            NodeOrToken::Token(it) => it.text_len(),
+            NodeOrToken::Token(it) => it.text_len_including_trivia(),
```
ChayimFriedman2, line 77: And here too.

src/green/node.rs
```diff
@@ -206,7 +206,7 @@ impl GreenNode {
         let mut text_len: TextSize = 0.into();
         let children = children.into_iter().map(|el| {
             let rel_offset = text_len;
-            text_len += el.text_len();
+            text_len += el.as_deref().text_len();
```
ChayimFriedman2, line 209: But then you need to change here.

src/green/token.rs
```diff
@@ -95,29 +109,79 @@ impl GreenTokenData {
...
-    /// Returns the length of the text covered by this token.
+    /// Returns the length of the text covered by this token, excluding its trivia.
     #[inline]
     pub fn text_len(&self) -> TextSize {
         TextSize::of(self.text())
     }
+
+    #[inline]
+    pub fn text_range(&self) -> TextRange {
+        TextRange::at(self.leading_trivia_len(), self.text_len())
```
ChayimFriedman2, line 128: This is O(N) for a hot operation. The `offset` of the red node should instead be the absolute offset (including the trivia).

src/cursor.rs
```diff
@@ -648,6 +665,17 @@ impl SyntaxToken {
...
+            return if self.text_range().start() < owner.text_range().start() {
+                owner.leading_trivia().nth(next).or(Some(owner))
+            } else {
+                owner.trailing_trivia().nth(next).or_else(|| {
+                    let after = owner.next_token()?;
+                    Some(after.leading_trivia().next().unwrap_or(after))
+                })
+            };
+        }
         match self.next_sibling_or_token() {
```
ChayimFriedman2, line 668: This will return the next token instead of its leading trivia, which is not correct. Same for `prev_token()`.

src/cursor.rs
```diff
@@ -798,12 +887,14 @@ impl fmt::Debug for SyntaxNode {
 
 impl fmt::Display for SyntaxNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.preorder_with_tokens()
-            .filter_map(|event| match event {
-                WalkEvent::Enter(NodeOrToken::Token(token)) => Some(token),
-                _ => None,
-            })
-            .try_for_each(|it| fmt::Display::fmt(&it, f))
+        let (Some(first), Some(last)) = (self.first_token(), self.last_token()) else {
```
ChayimFriedman2, line 853: Why this dance? This should print trivia, which makes this as simple as `self.text().fmt(f)`.

src/cursor.rs
ChayimFriedman2: This will also not be correct when we need to move to the first trailing trivia.

ChayimFriedman2 (changes requested):

src/cursor.rs
bit-aloo: Should text_range in token should include trivia or not? Because for node we do contain trivia for text_range but for the token we don't. I find this (including trivia text range), more coherent.

src/green/element.rs
bit-aloo: Same for this, I find this more coherent.

src/cursor.rs
bit-aloo: Ah!! shoot, none of this should come up. next_token and prev_token should only be used to traverse the canonical tree, not trivia. For trivia, we should get the leading/trailing trivia from the token directly and then continue traversing from there. My bad

src/cursor.rs
bit-aloo: Wait a second, is your thought process is like, text range should include the text owned by the node/token. In case of node it is whatever its token and trivia include, whereas for token its just its text not the trivia? That can be a reasonable way to look at it.

src/green/token.rs
bit-aloo: Should be O(1) now

src/cursor.rs
bit-aloo: removed it, Thanks

src/cursor.rs
ChayimFriedman2: > Should text_range in token should include trivia or not? Because for node we do contain trivia for text_range but for the token we don't. I find this (including trivia text range), more coherent.

That's one way to frame that. But another way is your second comment. Yet another way is that we preserve the existing behavior.

src/cursor.rs
ChayimFriedman2: You didn't seem to remove it?

src/cursor.rs
ChayimFriedman2: No, actually the opposite: `[prev/next]_token()` are heavily used in multiple places (in r-a) to visit the absolute next token in tree in order of tokens. Unlike siblings they do not consider different parent nodes to matter and they should also not consider trivia to not be a normal token. This implementation (now changed) was correct, I was pointing that it wasn't enough because while it correctly handled the case of calling `[next/prev]_token()` on a trivia token, it didn't correctly handle the case of calling `next_token()` on a non-trivia token where the next non-trivia token has leading trivia, or `prev_token()` on a non-trivia token where the prev non-trivia token has trailing trivia.

src/cursor.rs
ChayimFriedman2: Like I said I'm not entirely sure but this does feel better to me.

src/cursor.rs
bit-aloo: I should take some sleep. Will update soon

src/cursor.rs
bit-aloo: Any reason for that? like how see this?

ChayimFriedman2: What do you mean? I presented two reasons: this is actually more coherent if we view this as each element giving its "owned" range, and this is closer with the existing behavior.

bit-aloo: Updated.

src/green/element.rs
bit-aloo: Updated.

src/cursor.rs
bit-aloo: Fixed.

src/cursor.rs
bit-aloo: Removed again. This time fully awake :smile:

ChayimFriedman2 (approved):
Thanks!

I'm merging this but not releasing yet because we might want to make more adjustments to shrink the diff of the r-a PR, for example not printing trivia when empty like I said.

PR #220, opened 2026-09-08
More trivia changes

bit-aloo:


src/green/token.rs
```diff
@@ -152,6 +152,21 @@ impl GreenTokenData {
     pub(crate) fn text_len_including_trivia(&self) -> TextSize {
         self.leading_trivia_len() + self.text_len() + self.trailing_trivia_len()
     }
+
+    #[must_use]
+    pub fn splice_trivia<R, I>(&self, side: TriviaSide, range: R, replace_with: I) -> GreenToken
```
ChayimFriedman2: I'm hesitant to offer such API because new rowan cannot have it; in it, the only way to create a new node or token is a from-scratch builder. So if you want this to aid with the migration I'll happily merge it, but it cannot be the final form.

src/ast.rs
```diff
@@ -57,7 +57,7 @@ pub struct SyntaxNodePtr<L: Language> {
 impl<L: Language> SyntaxNodePtr<L> {
     /// Returns a [`SyntaxNodePtr`] for the node.
     pub fn new(node: &SyntaxNode<L>) -> Self {
-        Self { kind: node.kind(), range: node.text_range() }
+        Self { kind: node.kind(), range: node.text_range_without_outer_trivia() }
```
ChayimFriedman2: Why is `SyntaxNodePtr`, of all things, using this method? It could maybe be useful for other things but `SyntaxNodePtr` only purpose in life is locating nodes; the range doesn't matter as long as it's unique, so why choose the less natural range?

src/api.rs
```diff
@@ -280,6 +299,30 @@ impl<L: Language> SyntaxToken<L> {
         self.raw.index()
     }
 
+    pub fn child_index(&self) -> Option<usize> {
```
ChayimFriedman2: We should probably remove `index()` and call this `index()` instead. What `index()` does with trivia is highly non-natural and I wouldn't even want to commit to it.

src/api.rs
```diff
@@ -193,6 +204,14 @@ impl<L: Language> SyntaxNode<L> {
         self.raw.last_token().map(SyntaxToken::from)
     }
 
+    pub fn next_structural_token(&self) -> Option<SyntaxToken<L>> {
```
ChayimFriedman2: "Structural" isn't really descriptive. I suggest `[next/prev]_non_trivia_token()`. It's long but crystal clear.

src/cursor.rs
```diff
@@ -441,10 +515,24 @@ impl SyntaxNode {
     }
 
     pub fn first_token(&self) -> Option<SyntaxToken> {
-        self.first_child_or_token()?.first_token()
+        self.children_with_tokens().find_map(|child| child.first_token())
```
ChayimFriedman2: This doesn't include trivia, right? I missed this in the previous PR but we might want it to do.

src/cursor.rs
```diff
@@ -694,55 +799,76 @@ impl SyntaxToken {
...
+            (Some(left), Some(right)) => {
+                !left.is_trivia()
+                    && !right.is_trivia()
+                    && left.next_structural_token().as_ref() == Some(right)
+            }
+            (Some(left), None) => !left.is_trivia() && left.next_structural_token().is_none(),
+            (None, Some(right)) => !right.is_trivia() && right.prev_structural_token().is_none(),
+            (None, None) => false,
+        };
+        adjacent.then(|| {
```
ChayimFriedman2: Oh this only works if they are adjacent. So at least give it a name like `trivia_between_adjacent()`, and consider making it assert they're adjacent.

src/green/token.rs
bit-aloo: Moved it to r-a

src/api.rs
bit-aloo: Changed

src/api.rs
bit-aloo: Done

src/cursor.rs
bit-aloo: Nope, we have new set of APIs now.

src/cursor.rs
bit-aloo: This was horrible. I have new trivia_before and trivia after APIs exposed.

src/ast.rs
bit-aloo: Yup agree. Removed it.

src/cursor.rs
```diff
@@ -441,10 +510,41 @@ impl SyntaxNode {
     }
 
     pub fn first_token(&self) -> Option<SyntaxToken> {
```
ChayimFriedman2, line 512: The `first_X_token()` methods are now incorrect; they should return the absolute first token in the node, including indirect descendants.

ChayimFriedman2: And same for `last_X_token()`.

ChayimFriedman2 (changes requested):
One thing then LGTM.

src/cursor.rs
bit-aloo: These do return the absolute first/last token, first_non_trivia_token finds the first non-trivia token at any depth. first_token then returns that token's first leading trivia token, or the token itself if it has none. last_token works the same way with trailing trivia. Or I am missing something to understand?

ChayimFriedman2: Ah right, the `find_map()` confused map, but it's to handle empty nodes (with no tokens).

ChayimFriedman2:
Thanks! I'll release this as a prerelease.

ChayimFriedman2 (approved):

ChayimFriedman2:
Published `0.18.0-alpha.1`.

